PR #2, opened 2022-09-03
Add task configuration schema and describe client behaviour

wangshan:
First update, it's mostly a brain dump, please have a look. I'll fix the build failure early next week.


draft-wang-ppm-dap-taskprov.md
```diff
@@ -68,31 +72,165 @@ many deployments.
...
-> succinct and descriptive is ideal. `task_prov` is fine, but we can maybe do
+> succinct and descriptive is ideal. `task-prov` is fine, but we can maybe do
 > better.
 
 A new extension is defined:
 
 ~~~
 enum {
-   task_prov(0xff00), (65535)
+  task-prov(0xff00),
```
cjpatton, line 94: nit: I think 4-character tab is a bit more readable.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -68,31 +72,165 @@ many deployments.
...
+  (65535)
 } ExtensionType;
 ~~~
+> NOTE(shan) TLS uses underscore `_` to define codepoints, which is different
+> from DAP's convension of using hyphen `-`, which one to follow?
```
cjpatton, line 99: Good catch. I think underscore is better, but consistency with DAP is more important. I filed https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/314 to figure this out.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -68,31 +72,165 @@ many deployments.
...
 
 When the Client includes this extension with its report, the body of the
 extension is structured as follows:
 
 ~~~
-TODO(wangshan) Specify the structure of the extension.
+/* Definition of all parameters in extension_data for one task */
+struct {
+  /* Info specific for a task. */
+  opaque task_info<1..2^8-1>,
```
cjpatton, line 108: Here and below.

```suggestion
  opaque task_info<1..2^8-1>;
```

cjpatton (commented):
First pass of the schema. one high-level comment. The structure of the VDAF config is incomplete. Plus,  I think it could be spelled out cleaner. I imagine it looking something like this:

```
VdafType uint32_t; // Codepoint from I.D-draft-irtf-cfrg-vdaf-03

struct {
    VdafType type;
    select type (type) {
        Prio3Aes138Count: Empty;
        Prio3Aes128Sum: uint8_t bits;
        Prio3Aes128Histogram: uint64_t buckets<0..2^16-1>;
        Poplar1Aes128:  uint16 bits;
    }
} VdafConfig;
```

Then replace `vdaf_type` and `vdaf_parameter` with a single `VdafConfig`.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,43 +60,231 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+    /* A unique identifier for the VDAF instance used for the task, */
+    /* including the type of measurement associated with the task. */
+    VdafType vdaf_type;
+
+    /* Additional parameters relevant for the vdaf_type */
+    opaque vdaf_data<1..2^16-1>;
+} TaskConfig;
+
+/* Defined in DAP core protocol. */
```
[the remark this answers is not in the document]
wangshan: is `//` a valid comment format for TLS?

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,43 +60,231 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+enum {
+    reserved(0), // Reserved for testing purposes
+    time-interval(1),
+    fixed-size(2),
+    (65535)
+} QueryType;
+
+struct {
+    uint32 max_batch_size;
+} FixedSizeQueryConfig;
```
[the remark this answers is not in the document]
wangshan: wouldn't it be more clear if we have an explicitly type for it? for e.g. if we have another field required by FixedSizeQueryConfig in the future

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,43 +60,231 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+enum {
+    reserved(0), // Reserved for testing purposes
+    float16(1),
+    float32(2),
+    float64(3),
+    fixed-point16(4),
+    fixed-point32(5),
+    fixed-point64(6),
+    (7)
```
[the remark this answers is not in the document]
wangshan: type removed

wangshan:
@cjpatton 

```
VdafType uint32_t; // Codepoint from I.D-draft-irtf-cfrg-vdaf-03

struct {
    VdafType type;
    select type (type) {
        Prio3Aes138Count: Empty;
        Prio3Aes128Sum: uint8_t bits;
        Prio3Aes128Histogram: uint64_t buckets<0..2^16-1>;
        Poplar1Aes128:  uint16 bits;
    }
} VdafConfig;
```
why do we need two definitions of `VdafType uint32_t` and `VdafType type`, is the former a type alias?

I'm thinking the VdafConfig struct is up to the VDAF, making it opaque will allow new VDAFs to be defined without changing this extension draft (like an extension on this extension), otherwise wouldn't this extension be tied to all the VDAFs above?

draft-wang-ppm-dap-taskprov.md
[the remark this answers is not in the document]
cjpatton: It's just my opinion, but I don't think this is more clear, since you now have to look in two places to understand the information conveyed by QueryConfig. However, defining new structs is sometimes appropriate. For example:
1.  when either you need to serialize the same data in multiple places in the protocol
2. when there are a large number of parameters in the struct.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+    opaque task_info<1..2^8-1>;
+
+    /* A list of URLs relative to which an aggregator's API endpoints */
+    /* can be found. */
+    Url aggregator_endpoints<1..2^16-1>; // Defined in I-D.draft-ietf-ppm-dap-02
+
+    /* This determines the query type for batch selection and the */
+    /* properties that all batches for this task must have. */
+    QueryConfig query_config; // Defined in I-D.draft-ietf-ppm-dap-02
+
```
cjpatton, line 157: nit (weird mix of block and inline comments; similarly below): 
```suggestion
    /* properties that all batches for this task must have.  Defined in I-D.draft-ietf-ppm-dap-02. */
    QueryConfig query_config;

```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -24,9 +24,13 @@ venue:
...
  -
-    fullname: Your Name Here
-    organization: Your Organization Here
+    fullname: Shan Wang
+    organization: Apple Inc.
     email: "shan_wang@apple.com"
+ -
+    fullname: Christopher Patton
+    organization: Cloudflare
+    email: "cpatton@cloudflare.com"
```
cjpatton, line 33:
```suggestion
    email: "chrispatton+ietf@gmail.com"
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -140,3 +379,23 @@ TODO(wangshan) Specify the structure of the extension.
...
+Apple Inc.
+mscaria@apple.com
+
+Suman Ganta
+Apple Inc.
+sganta2@apple.com
+
+Christopher A. Wood
+Cloudflare
+chriswood@cloudflare.com
```
cjpatton, line 400: Chris uses his personal email for IETF contributions
```suggestion
caw@heapingbits.net
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+    opaque vdaf_data<1..2^16-1>;
+} TaskConfig;
+
+struct {
+    uint32 max_batch_size;
+} FixedSizeQueryConfig;
+
+struct {
+    QueryType query_type; // Defined in I-D.draft-ietf-ppm-dap-02
+    Duration time_precision; // Defined in I-D.draft-ietf-ppm-dap-01
```
cjpatton, line 180:
```suggestion
    QueryType query_type;    // Defined in I-D.draft-ietf-ppm-dap-02
    Duration time_precision; // Defined in I-D.draft-ietf-ppm-dap-02
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+    /* This determines the query type for batch selection and the */
+    /* properties that all batches for this task must have. */
+    QueryConfig query_config; // Defined in I-D.draft-ietf-ppm-dap-02
+
+    /* The maximum number of times a batch of reports may be queried */
+    /* by the Collector. */
+    uint16 max_batch_lifetime;
+
+    /* Time up to which clients are allowed to upload to this task. */
```
cjpatton, line 162: nit: In the VDAF spec I've adopted the convention of capitalizing Client, Aggregator, and Collector. For the DAP spec I've been pushing for doing the same. We can follow this convention or not, but we should try to be consistent.
```suggestion
    /* Time up to which Clients are allowed to upload to this task. */
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
 extension is structured as follows:
 
 ~~~
-TODO(wangshan) Specify the structure of the extension.
+struct {
+    /* Info specific for a task. */
+    opaque task_info<1..2^8-1>;
+
+    /* A list of URLs relative to which an aggregator's API endpoints */
+    /* can be found. */
```
cjpatton, line 151: Probably best to follow block comment conventions in the DAP spec.
```suggestion
    /* A list of URLs relative to which an aggregator's API endpoints
     can be found. */
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+    /* The maximum number of times a batch of reports may be queried */
+    /* by the Collector. */
+    uint16 max_batch_lifetime;
+
+    /* Time up to which clients are allowed to upload to this task. */
+    /* See https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/304 */
+    Time task_expiration; // Defined in I-D.draft-ietf-ppm-dap-02
+
+    /* A unique identifier for the VDAF instance used for the task, */
```
cjpatton, line 166: We need to say here that this is a codepoint for either a standard VDAF or reserved for private use.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+    /* including the type of measurement associated with the task. */
+    VdafType vdaf_type;
+
+    /* Additional parameters relevant for the vdaf_type */
+    opaque vdaf_data<1..2^16-1>;
```
cjpatton, line 171:
```suggestion
    opaque vdaf_config<1..2^16-1>;
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+    /* Time up to which clients are allowed to upload to this task. */
+    /* See https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/304 */
+    Time task_expiration; // Defined in I-D.draft-ietf-ppm-dap-02
+
+    /* A unique identifier for the VDAF instance used for the task, */
+    /* including the type of measurement associated with the task. */
+    VdafType vdaf_type;
+
+    /* Additional parameters relevant for the vdaf_type */
```
cjpatton, line 170:
```suggestion
    /* Additional parameters relevant for the vdaf_type. */
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+* An opaque `vdaf_data` that contains any VDAF specific parameters for the
+chosen `vdaf_type`. The aggregators MUST pass `vdaf_data` to VDAF initialiser,
+based on the chosen `vdaf_type`.
+
+The definition of Time, Duration, Url, QueryType follow those in
+{{?DAP=I-D.draft-ietf-ppm-dap-01}}.
+
+Note that the parameters that are not necessarily tied to a task (for e.g.
+collector hpke config), and secrets that should not be known by clients, like
+`vdaf_verify_key`, MUST still be exchanged out-of-band among aggregators.
```
cjpatton, line 208: I don't think this paragraph adds anything. It can be removed.

cjpatton (commented):
Extension structure is looking good. I'll look at the rest in the next pass.

> ```
> VdafType uint32_t; // Codepoint from I.D-draft-irtf-cfrg-vdaf-03
> 
> struct {
>     VdafType type;
>     select type (type) {
>         Prio3Aes138Count: Empty;
>         Prio3Aes128Sum: uint8_t bits;
>         Prio3Aes128Histogram: uint64_t buckets<0..2^16-1>;
>         Poplar1Aes128:  uint16 bits;
>     }
> } VdafConfig;
> ```
> 
> why do we need two definitions of `VdafType uint32_t` and `VdafType type`, is the former a type alias?

* `VdafType uint32_t;` defines `VdafType` as a type alias for `uint32_t`.
* `VdafType type;`, in the given context, defines the type of a field struct. It does not define a type. Instead, what it says is that the struct includes a `VdafType`.


> I'm thinking the VdafConfig struct is up to the VDAF, making it opaque will allow new VDAFs to be defined without changing this extension draft (like an extension on this extension), otherwise wouldn't this extension be tied to all the VDAFs above?

That's a good point: We need to make sure that unspecified VDAFs can still be encoded. This seems like a reasonable way to do it. However, for this extension to be maximally useful, it will be necessary to spell out serialization of the configs of standardized VDAFs. Just something to keep in mind for the future.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+`ReportMetadata`'s extenion field, and save the encoded `TaskConfig` in
+`extension_data` field.
 
 # Leader Behavior
 
 > TODO(wangshan) Say what the Leader does when it receives a report with this
 > extension. What do Leaders do if they *don't* support this extension? Is it
 > safe to ignore it? How are the upload, aggregate, and collect flows impacted?
 
+Leader should have saved any information that is not always tied to a
```
junyechen1996, line 258: I wonder if we should briefly mention all the "secret" parameters in a task that cannot be agreed through the extension-based task provisioning. On top of my head, these include `vdaf_verify_key`, `collector_hpke_config`, and `aggregator_auth_tokens`

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+Upon receipt of a `AggregateInitializeReq`, helper reads the extension
+codepoint in `extension_type`. If helper does not support this extension, it
+SHOULD abort the aggregate protocol and alert the leader with error
+"unrecognizedMessage".
+
+If helper supports `task-prov` extension, it should first check if the task ID
+already exists, if so helper continues to the rest of helper initialization as
+usual (see helper initialization in {{?DAP=I-D.draft-ietf-ppm-dap-01}}.)
+Note if the existing tasks's configuration is different from the one in report
```
junyechen1996, line 326: existing tasks's -> existing task's

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+type `unrecognizedTask`, if it supports this extension, then it should retry
+with the same `CollectReq`, potentially using an interval from the Retry-After
+header in the last `CollectResp`.
+
+# Operational Considerations
+
+> NOTE(shan) Do we want to include this section in the beginning?
+
+The in-band task provision mechanism is easy to implement with streaming
+framework that has `groupBy operator`. In fact, task as an object doesn't have
```
junyechen1996, line 354: `groupBy operator` -> `groupBy` operator

junyechen1996 (commented):
First pass mostly for spelling mistakes and wordings.

draft-wang-ppm-dap-taskprov.md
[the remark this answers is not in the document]
wangshan: I'll change this since I don't know what else should be include in FixedSizeQueryConfig yet.

draft-wang-ppm-dap-taskprov.md
wangshan: @cjpatton The core protocol task configuration includes secrets like `vdaf_verify_key`, I think this should be mentioned explicitly?  @junyechen1996 your comment https://github.com/wangshan/draft-wang-ppm-dap-taskprov/pull/2/files#r964267738 is basically addressed here

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
 
+The client should know whether `task-prov` extension will be used and all
+parameters required for `TaskConfig` prior to constructing the extension
+body, either out-of-band from aggregators, or from information already saved on
+client.
+
+To use this extension, the client should first decide what is the task ID, see
+{{construct-task-id}}.
+
+## Construct task ID {#construct-task-id}
```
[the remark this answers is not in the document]
wangshan: yes, but thinking about this more, we probably should enforce task ID to be created involving task config, otherwise the aggregator might aggregate the same task ID with different TaskConfig.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,304 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+serves as an unique byte array to distinguish different tasks that share the
+same task parameters. Task ID can be constructed on server side and then
+distributed to clients, or constructed independently on clients and servers.
+When constructed independently, the mechanism used for creating the task ID
+must be known to both clients and the collector.
+
+## Construct extension body
+
+Client constructs this extension during the upload flow, after hpke config
+(see update flow and hpke-config in {{?DAP=I-D.draft-ietf-ppm-dap-01}}.). Note
```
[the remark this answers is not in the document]
wangshan: yes that's pretty much it, let me remove this sentence

draft-wang-ppm-dap-taskprov.md
wangshan: that is mentioned just above the client behaviour section

wangshan:
> That's a good point: We need to make sure that unspecified VDAFs can still be encoded. This seems like a reasonable way to do it. However, for this extension to be maximally useful, it will be necessary to spell out serialization of the configs of standardized VDAFs. Just something to keep in mind for the future.

I'll bring back the standardized VDAFs in a VdafConfig.

draft-wang-ppm-dap-taskprov.md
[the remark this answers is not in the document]
cjpatton: Yes

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,302 @@ collector). Thus it is up to each deployment of DAP to securely implement its
 own mechanism.
 
 This document describes a mechanism for configuring tasks that may be useful in
-many deployments.
+many deployments. The goal of this mechanism is to add transparency to the
```
cjpatton, line 63: I continue to think that "transparency" is not the right sales pitch for this extension. The reason is this: To sell it this way, you'll need to (1) define transparency and (2) to make the case that vanilla DAP *isn't* transparent. So far I don't see a clear/useful definition of transparency for which this case can be made.

Suppose we define transparency this way: "A DAP deployment is transparent if what is being measured, and how, is evident to the client." Assuming this definition, it's not hard to argue that draft-wang-ppm-taskprov does not add any value relative to vanilla DAP. The important question is: What information does the client have now that it didn't have before? Quoting from the original issue (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/290):

> By sending all task parameters to clients then return to the server, we add transparency to the task that clients participate in Client has the choice to see what data has been collected with what parameters (like min_batch_size, max_batch_lifetime, or any differential privacy parameters if that's the privacy guarantee used).  For some clients these task parameters can even be hardcoded on the client side (for e.g. on a mobile device) to avoid any tempering from server.

Why is this not true for vanilla DAP? It seems like a deployment is free to surface this information to clients, regardless of how the task is configured. In fact, at some point the client software needs this information in order to participate in the protocol. It's also possible to hardcode DP parameters and use them for multiple tasks for vanilla DAP.

On the other hand, I think a *very* strong case for this extension can be made without talking about transparency at all. Namely, it specifies a means of configuring tasks without intervention of a third-party; all information necessary for aggregating a report is carried by the report itself. This is an elegant idea on its own.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,302 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+task provision process, and define a task provision method that does not rely
+on deployment specific leader-helper out-of-band agreement.
```
cjpatton, line 65: Vanilla DAP does not mandate some sort of "leader-helper" agreement. Rather, the means by which parties are configured is not specified.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,302 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+2. No need for out-of-band task orchestration between leader and helpers, any
+leaders and helpers can work on the same DAP task, therefore making adoption of
+DAP easier.
+
+3. A "task" now means the same task ID and same task configuration, including
+parameters for choosing VDAF. Therefore, reports from one task will only be
+aggregated with one VDAF. For the same reason, if a malicious client changes
+the task ID or task configuration, its report will be aggregated in a different
+task, with other poison reports from the same malicious attack. The "good"
+reports will not be polluted.
```
cjpatton, line 93: I don't think this property (3.) is new to draft-wang-ppm-taskprov. In fact, it ought to hold for any deployment of DAP.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,302 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
->
-> - Why does this mechanism need to be an extension (i.e., why are protocol
->   changes needed), as opposed to something that happens completely
->   out-of-band?
->
-> - Which of the sub-protocols are changed: upload, aggregate, collect?
+Because extension is used in HPKE's AAD, by including all task configurations
+in the extension, malicious leader cannot change the task configuration to
+mislead helpers to reduce privacy guarantee, for e.g. by reducing
+`min_batch_size`.
```
cjpatton, line 98: This property as well should hold for any deployment of DAP. Since it's not particular to this extension, I don't think it's worth mentioning in the intro. Of course, in Security Considerations we'll want to discuss why this extension preserves the properties of DAP (i.e., does not introduce any regressions to security).

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,302 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
 
+Similar to {{?DAP=I-D.draft-ietf-ppm-dap-01}}, this document uses the verbs
+"abort" and "alert with `[some error message]`" to describe how protocol
+participants react to various error conditions.
 
-# The "task_prov" Extension
+The terms used follow those described in {{?DAP=I-D.draft-ietf-ppm-dap-01}}. The
+following new terms are used:
+
+Task provision:
```
cjpatton, line 113:
```suggestion
Task provisioning:
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,302 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+     I-D.draft-ietf-ppm-dap-02. */
+    QueryConfig query_config;
+
+    /* The maximum number of times a batch of reports may be queried by
+    the Collector. */
+    uint16 max_batch_lifetime;
+
+    /* Time up to which Clients are allowed to upload to this task. See
+     https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/304. Defined
+     in I-D.draft-ietf-ppm-dap-02. */
```
cjpatton, line 164:
```suggestion
    /* Time up to which Clients are allowed to upload to this task. See
    https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/304. Defined
    in I-D.draft-ietf-ppm-dap-02. */
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,302 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+    Time task_expiration;
+
+    /* A codepoint for either a standard VDAF or reserved for private
+     use. */
```
cjpatton, line 168:
```suggestion
    /* A codepoint defined in I-D.draft-irtf-cfrg-vdaf-03 or reserved for private
    use. */
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,302 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+    uint32 min_batch_size;
+    select (query_type) {
+        case time-interval: Empty;
+        case fixed-size: uint32 max_batch_size;
+    }
+} QueryConfig;
 ~~~
 
+The purpose of `TaskConfig` is to include all parameters that are necessary
+for creating a new task in aggregators. It includes all the fields to be
```
cjpatton, line 187:
```suggestion
The purpose of `TaskConfig` is to define all parameters that are necessary
for configuring an aggregator. It includes all the fields to be
```

draft-wang-ppm-dap-taskprov.md
cjpatton: Hmm, I don't think my comment was very helpful. Please allow me to try again :)

It's useful to note that TaskConfig does not encode all parameters required to run a task. (Listing them is also helpful: I believe this just includes the VDAF verification key and the collector HPKE config.)

What's less useful here is saying that these MUST be exchanged out-of-band. It's best to reserve the keyword "MUST" for prescriptive things, e.g., "If the client does X, then the server MUST do Y". Here we're saying that there are fields that MUS be exchanged out-of-band, but we don't say how.

I would be happy with something like this: 

> Note that `TaskConfig` does not encode all of the parameters required for the aggregator to run a task. In particular, the VDAF verification key, the collector HPKE configuration, and whatever assets are required for HTTP request authentication are still established out-of-band.

Side note: Is the plan to to derive the VDAF verification key from the a shared secret on-the-fly?

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,302 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+of task configuration, as long as the generated task ID is determinstic and
+stay consistent across all parties. For example, the task ID can be a SHA256
+hash of the entire serialized `TaskConfig`. In this case, the `task_info`
+serves as an unique byte array to distinguish different tasks that share the
+same task parameters. Task ID can be constructed on server side and then
+distributed to clients, or constructed independently on clients and servers.
+When constructed independently, the mechanism used for creating the task ID
+must be known to both clients and the collector.
+
+> OPEN ISSUE: Should task ID construction from TaskConfig be enforeced?
```
cjpatton, line 239: I think so. If the goal of this extension is to do as much in-band provisioning as possible, then permitting flexibility here defeats the purpose because it requires the parties to agree, out-of-band, on the method for choosing the batch ID.

As has been pointed out, diversifying the task ID is important. One way to do this is to rely on the `task_info` to be unique across the entire deployment. Alternatively, we could add a nonce field to the TaskConfig, which the task author should fill with random bytes.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,69 +60,302 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+## Construct extension body
```
cjpatton, line 241: This subsection is not necessary, it's sufficient to just define the extension.

cjpatton (changes requested):
Pass of intro and client behavior.

draft-wang-ppm-dap-taskprov.md
wangshan: let me move this to Security Considerations as a quoted comment, we can update/remove it when Security Considerations are fleshed out.

draft-wang-ppm-dap-taskprov.md
wangshan: I see, let me update it.

Re: vdaf_verify_key, I think our last conclusion was that if the leader chooses vdaf_verify_key, then the privacy leak threat already exists, generate it using a shared secret doesn't help, do we have consensus on this?

draft-wang-ppm-dap-taskprov.md
wangshan: I agree with the part that transparency in the form of "client knows the parameters" is true to vanilla DAP. 

But what this extension adds is: the client's measurement will be aggregated using the parameters, because aggregators use these exact parameters to define a task. While in vanilla DAP, there's nothing stopping task authors from giving clients and helpers different parameters.

draft-wang-ppm-dap-taskprov.md
wangshan: agreed. But I think the part that filters out polluted reports is still worthy calling out, let me move it to Implementation and Operational Considerations, I can delete it if you think that's clearer.

draft-wang-ppm-dap-taskprov.md
cjpatton: That's definitely a useful property of this extension, and it should be called out in the intro. But I don't think it has to do with transparency.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,333 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+`VdafConfig` is not specified in this document, instead it should be defined by
+each VDAF implementation, for example, the simplest VDAF config for Prio3 can
+be defined as:
+
+~~~
+struct {
+    select (vdaf_type) {
+        case Prio3Aes128Count: Empty;
+        case Prio3Aes128Sum: uint8 bits;
+        case Prio3Aes128Histogram: uint32 buckets;
```
cjpatton, line 212: This should actually be a list of buckets. Plus, uint64_t is more useful for defining bucket boundaries.
```suggestion
        case Prio3Aes128Histogram: uint64 buckets<8, 2^16-8>;
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,333 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+enum {
+    Prio3Aes128Count(0x00000000),
+    Prio3Aes128Sum(0x00000001),
+    Prio3Aes128Histogram(0x00000002),
+    Poplar1Aes128(0x00001000),
+    (2^32-1)
+} VdafType;
+~~~
+
+`VdafConfig` is not specified in this document, instead it should be defined by
```
cjpatton, line 203:
```suggestion
The structure of the `vdaf_config` field is not specified in this document, instead it should be defined by
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,333 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+each VDAF implementation, for example, the simplest VDAF config for Prio3 can
+be defined as:
```
cjpatton, line 205: If we're going to spell out the structure for Prio3, we should be more declaritive.

> For VDAFs specified in {{!VDAF}}, implementations SHOULD use the following structure:

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,333 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+chosen `vdaf_type`. The aggregators MUST pass `vdaf_config` to VDAF initialiser,
+based on the chosen `vdaf_type`.
+
+The codepoints for standardized (V)DAFs are listed below:
+
+~~~
+/* Codepoint for each standardized VDAF. Defined in
+ I-D.draft-irtf-cfrg-vdaf-02 */
+enum {
+    Prio3Aes128Count(0x00000000),
```
cjpatton, line 195: Here and below (and update VdafConfig to match)
```suggestion
    prio3-aes128-count(0x00000000),
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,333 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+    select (vdaf_type) {
+        case Prio3Aes128Count: Empty;
+        case Prio3Aes128Sum: uint8 bits;
+        case Prio3Aes128Histogram: uint32 buckets;
+    }
+} VdafConfig;
+~~~
+
+> OPEN ISSUE: Should DP parameters be defined as a different "dimension" to
+> VDAF, given that it's likely various DP mechanisms can be applied to any VDAF.
```
cjpatton, line 218: That probably makes since, but it somewhat depends on the outcome of https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/94. Let's keep this issue open and add a reference to the VDAF issue.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,333 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+The definition of Time, Duration, Url, QueryType follow those in
+{{?DAP=I-D.draft-ietf-ppm-dap-01}}.
+
+Note that `TaskConfig` does not encode all of the parameters required for the
+aggregator to run a task. In particular, parameters that should not be known to
+clients, like the VDAF verification key, the collector HPKE configuration, and
+whatever assets are required for HTTP request authentication are still
+established out-of-band.
```
cjpatton, line 227: I think this extension should spell out derivation of the VDAF verification key, as this needs to be unique per task. Feel free to leave as an OPEN ISSUE in this PR.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,333 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+client.
+
+To use this extension, the client should first decide what is the task ID, see
+{{construct-task-id}}.
+
+Client constructs this extension during the upload flow, after hpke config
+(see update flow and hpke-config in {{?DAP=I-D.draft-ietf-ppm-dap-01}}.).
+Client typically sets `extension_type` to `task-prov` codepoint in
+`ReportMetadata`'s extension field, and save the encoded `TaskConfig` in
+`extension_data` field.
```
cjpatton, line 246:
```suggestion
To offer the "task-prov" extension, the client adds the `TaskConfig` structure it received from the task author in the extensions field of its `Report`. It computes the task ID as described in {{construct-task-id}}.
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,333 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+## Construct task ID {#construct-task-id}
+
```
cjpatton, line 249: > Task ID can be constructed
from `task_info` and any other part of task configuration, as long as the
generated task ID is determinstic and stay consistent across all parties. For
example, the task ID can be a SHA256 hash of the entire serialized
`TaskConfig`. 

The spec needs to be declaritive here. Instead of saying "here's one way one might compute the task ID", spell it out explicitly. For example:

> When the "task-prov" extension is used, the task ID is computed as follows:
> 
> ~~~
> task_id = SHA-256(task_config)
> ~~~

where `task_config` is the payload of the extension sent by the client.

cjpatton (changes requested):

wangshan:
@cjpatton why a prio3 histogram uses a list of buckets, is it for non-continuous buckets? does this mean the dimension of the one hot vector can only be (2^16-8)/8 = 8191

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,330 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+the decoded report, and save task configuration with the newly created task. In
+particular, leader should deserialize `vdaf_config` corresponding to `vdaf_type`,
+and pass the relevant parameters to the VDAF initializer. At this point, the
+task provision step has completed, and leader should continue to the rest of
+upload flow.
+
+Leader MAY return error to the client if task creation failed.
+
+> OPEN ISSUE: would returning error reveal the position of the client in the
+> batch of reports received? if so is this a security or privacy threat?
```
cjpatton, line 301: Can you elaborate more on "position of the client in the batch of reports"? Is there anything that the client learns from the error response that it couldn't compute on its own?

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,330 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+## Change to leader aggregate sub-protocol
+
+There is no change to the aggregate sub-protocol on leader side.
```
cjpatton, line 305: Then I think we can just get rid of this section.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,330 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+## Change to collect sub-protocol
+
+During collect init, leader must perform batch validation (see batch-validation
+in {{?DAP=I-D.draft-ietf-ppm-dap-01}}) to the `CollectReq`. This requires leader
+to know the parameters associated with the DAP task. However, upon receipt of
+`CollectReq`, leader may not know the parameters if the first upload has not
+arrived yet. In this case the leader MAY respond with HTTP status code 404
+Not Found and an error of type `unrecognizedTask`. The response MAY include a
+Retry-After header field to suggest a pulling interval to the collector.
```
cjpatton, line 315: Are changes actually needed? What's the goal here: to signal to the collector that the task hasn't been configured yet, but might eventually? Why is this useful?

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,330 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+> Alternatively the leader can just response with 303 and continue waiting.
 
 # Helper Behavior
 
-> TODO(wangshan) Say what the Helper does when it receives a report share with
-> this extension. (Same questions as above.)
+Helper should have received out-of-band from leader any information that is not
+always tied to a particular task, for e.g. `vdaf_verify_key`. Similar to
+leader, helper may not know the task ID of the current task before receiving
+the first `AggregateInitializeReq` message from the leader.
```
cjpatton, line 324: I think this spec should spell out how the vdaf verify key is computed. Say explicitily what else is out-of-band (HTTP request auth assets, collector HPKE config)

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,330 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+## Change to helper aggregate sub-protocol
 
+Upon receipt of a `AggregateInitializeReq`, helper reads the extension
+codepoint in `extension_type`. If helper does not support this extension, it
+SHOULD abort the aggregate protocol and alert the leader with error
+"unrecognizedMessage".
```
cjpatton, line 331: Here again, let's assume that the Helper MUST ignore unrecognized extensions and make sure we spell out the Helper behavior in a way that make sure ignore this is safe.

cjpatton, line 331: Again: Let's assume that the Helper MUST ignore unrecognized extensions. See https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/334

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,330 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+If the task ID has not been seen before, helper should read and decode report
+share's `extension_data` with the `TaskConfig` schema. If the decode failed,
+it MUST abort the aggregate protocol and alert the leader with error
+"unrecognizedMessage".
+
+If the decode succeeds, helper should create a new task using the task ID from
+the decoded report share, and save task configuration with the newly created
+task. In particular, helper should pass `vdaf_config` to the VDAF initializer,
+based on `vdaf_type` in `TaskConfig`. At this point, the task provision step has
+completed, and helper should continue to the rest of helper initialization.
```
cjpatton, line 348: Lots of this text is identical to the Leader's behavior. Can we try to de-duplicate it somehow? Perhaps we can add a subsection to the "Extension" section that describes this. Maybe call it "Provisioning a Task".

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,330 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
 > TODO(wangshan) Describe if/how this extension impacts Collector behavior.
 
+If Collector supports `task-prov` extension and receives a HTTP status code
+404 Not Found with error type `unrecognizedTask` after sending a `CollectReq`,
+it SHOULD retry with the same `CollectReq`, potentially using an interval from
+the Retry-After header in the received response.
+
+# Implementation and Operational Considerations
+
+> NOTE(shan) Do we want to include this section in the beginning?
```
cjpatton, line 361: Better to have it toward the end.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,330 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+A "task" now means the same task ID and same task configuration, if a malicious
+client changes the task ID or task configuration, its report will be aggregated
+in a different task, with other poison reports from the same malicious attack.
+The "good" reports will not be polluted.
```
cjpatton, line 366: this paragraph strikes me as a security consideration.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,330 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+The in-band task provision mechanism is easy to implement with streaming
+framework that has `groupBy` operator. In fact, task as an object doesn't have
```
cjpatton, line 369: > streaming framework that has `groupBy` operator.

This is too vague. If you want to suggest a particular implementation strategy, then you'll need to elaborate more here.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,330 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+to exist in aggregators, it mainly becomes an identifier to group aggregations
+together.
+
+This mechanism brings added overhead in `Report` and `ReportShare` since more
+duplicated data is passed around. Some optimisation can be done by sending only
+one copy of extension in `AggregateInitializeReq`.
```
cjpatton, line 375: I agree, changes to core DAP are likely warranted. However I think it's too early to suggest anything concrete, or even file an issue. I would make this paragraph an OPEN ISSUE.
```suggestion
> OPEN ISSUE: This mechanism brings added overhead in `Report` and `ReportShare` since more
> duplicated data is passed around. Some optimisation can be done by sending only
> one copy of extension in `AggregateInitializeReq`.
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,330 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
 # Security Considerations
 
 > NOTE(cjpatton) In this section we would describe any security goals we have
 > that go beyond the core DAP spec. We will also discuss if/how the extension
 > impacts the security of DAP itself.
 
+> Because extension is used in HPKE's AAD, by including all task configurations
+> in the extension, malicious leader cannot change the task configuration to
+> mislead helpers to reduce privacy guarantee, for e.g. by reducing
+> `min_batch_size`.
```
cjpatton, line 386: How does this compare with vanilla DAP? For this property to hold, it's sufficient to ensure a 1:1 mapping from task ID to task config. This is already required by vanilla DAP, so I don't actually think we gain anything here.

cjpatton (commented):
What should an Aggregator do if the task ID does not match the hash of the task config? We'll need to spell this out in both the "Leader Behavior" and "Helper Behavior" sections.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,330 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+By defining this mechanism as an extension, we can guarantee that for any
+deployments implementing this extension, all information required for
+aggregating a report is included in the report itself. There is no need for
+out-of-band task orchestration between leader and helpers, therefore making
+adoption of DAP easier.
 
+Implementations of this extension also has better privacy and security since
+aggregators must create tasks using the same parameters client sent. Task
+authors cannot give clients and aggregators different task parameters to reduce
+privacy guarantee on the aggregate result.
```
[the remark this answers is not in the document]
wangshan: makes sense, let me reword this

draft-wang-ppm-dap-taskprov.md
wangshan: I can't remember what I was worried about, let me delete this and add back later with example

draft-wang-ppm-dap-taskprov.md
wangshan: what does "ignore" mean in this case? it should carry on preparing the shares as if task ID is already associated with an existing task? The current core protocol doesn't specify what happens in the aggregate-flow if the task ID is not recognized. Maybe we should specify that behaviour in the core protocol first, then we can simply see "ignore" in this extension.

draft-wang-ppm-dap-taskprov.md
wangshan: I don't think this paragraph is essential, let me remove this from the first draft and add details later

cjpatton:
> @cjpatton why a prio3 histogram uses a list of buckets, is it for non-continuous buckets? does this mean the dimension of the one hot vector can only be (2^16-8)/8 = 8191

For this type, `buckets` defines the boundaries between the buckets: See https://www.ietf.org/archive/id/draft-irtf-cfrg-vdaf-03.html#name-prio3aes128histogram.

Yes, this would constrain the dimension of the vector. We could bump the length prefix to 24 or 32 bits if needed.

draft-wang-ppm-dap-taskprov.md
wangshan: this is necessary because batch-validation in core protocol relies on various task parameters known at the time of receiving collectReq, the goal is to let leader have a chance to let collector know what happened, otherwise, how does leader check for e.g. min_batch_size

wangshan:
> What should an Aggregator do if the task ID does not match the hash of the task config? We'll need to spell this out in both the "Leader Behavior" and "Helper Behavior" sections.

@cjpatton do we need to do anything? the current text doesn't specify any checks on aggregator side about task ID vs. hash(task_config). Any bad task ID will be grouped to it's own (bad) task.

wangshan:
> > @cjpatton why a prio3 histogram uses a list of buckets, is it for non-continuous buckets? does this mean the dimension of the one hot vector can only be (2^16-8)/8 = 8191
> 
> For this type, `buckets` defines the boundaries between the buckets: See https://www.ietf.org/archive/id/draft-irtf-cfrg-vdaf-03.html#name-prio3aes128histogram.
> 
> Yes, this would constrain the dimension of the vector. We could bump the length prefix to 24 or 32 bits if needed.

So this is related to this issue in DAP: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/311?
I'll create an issue in VDAF to bump this, 8192 is very limiting when used with differential privacy.

cjpatton:
> > > @cjpatton why a prio3 histogram uses a list of buckets, is it for non-continuous buckets? does this mean the dimension of the one hot vector can only be (2^16-8)/8 = 8191
> > 
> > 
> > For this type, `buckets` defines the boundaries between the buckets: See https://www.ietf.org/archive/id/draft-irtf-cfrg-vdaf-03.html#name-prio3aes128histogram.
> > Yes, this would constrain the dimension of the vector. We could bump the length prefix to 24 or 32 bits if needed.
> 
> So this is related to this issue in DAP: [ietf-wg-ppm/draft-ietf-ppm-dap#311](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/311)? I'll create an issue in VDAF to bump this, 8192 is very limiting when used with differential privacy.

This not related to https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/311. It's also not relevant to VDAF, which does not specify an encoding of VDAF config parameters. If you're expecting needing a large histogram, then all we need to do is increase the maximum length of the buckets list here.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,296 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+By defining this mechanism as an extension, we can guarantee that for any
+deployments implementing this extension, all information required for
+aggregating a report is included in the report itself. There is no need for
+out-of-band task orchestration between leader and helpers, therefore making
+adoption of DAP easier. Since aggregators must create tasks using the same
+parameters client sent, this extension prevents task authors from giving clients
+and aggregators different task parameters in order to achieve reduced privacy
+guarantee, without requiring an out-of-band mechanism.
+
+This extension affects upload, aggregate and collect sub-protocols.
```
cjpatton, line 82: This section should define the term "task author" (introduced by this draft) and elaborate on its role. We can do so in follow-up PR.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,296 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+instead it needs to be defined by the VDAF itself. For VDAFs specified
+in {{?VDAF=I-D.draft-irtf-cfrg-vdaf-01}}, implementations SHOULD use the
+following structure:
+
+~~~
+struct {
+    select (vdaf_type) {
+        case prio3-aes128-count: Empty;
+        case prio3-aes128-sum: uint8 bits;
+        case prio3-aes128-histogram: uint64 buckets<8, 2^16-8>;
```
cjpatton, line 204:
```suggestion
        case prio3-aes128-histogram: uint64 buckets<8, 2^24-8>;
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,296 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+
+To offer the `task-prov` extension, the client adds the `TaskConfig` structure
+it received from the task author in the extensions field of its `Report`. It
+computes the task ID as described in {{construct-task-id}}.
+
+## Construct task ID {#construct-task-id}
 
+For `task-prov` extension, a DAP task is not created before distributing task
+configuration to clients. Therefore, clients, aggregators and collector should
+construct a DAP task ID prior to uploading. A DAP task ID is computed as
```
cjpatton, line 244: Declarative
```suggestion
configuration to clients. Therefore, clients, aggregators and collector
construct the DAP task ID prior to uploading. A DAP task ID is computed as
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,296 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+If aggregator supports `task-prov` extension, it should first check if the
+task ID already exists, if so the aggregator continues to the rest of the flow
+being processed. Note if the existing tasks's configuration is different from
+the one in extension, HPKE decryption will fail due to mismatched AAD.
+
+If the task ID has not been seen before, aggregator should read and decode
+`extension_data` with the `TaskConfig` schema. If the decoding failed, it MUST
+abort the sub protocol with error "unrecognizedMessage".
+
+If the decoding succeeds, aggregator should create a new task using the task ID
```
cjpatton, line 265: Declaritive
```suggestion
If the decoding succeeds, aggregator creates a new task using the task ID
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,296 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+## Change to upload sub-protocol
+
+Upon receiving a report, leader reads the extension codepoint in
+`extension_type`. If the leader does not support this extension, it MUST ignore
+it. In particular, if the task ID is not known, then it MUST abort the handshake
+with "unrecognizedTask".
+
+> reason for aborting: if leader ignores it, then there is no way to ensure
+> clients that the extension is used and task parameters are respected in
+> aggregators
```
cjpatton, line 286: No need to include this paragraph.
```suggestion
```

draft-wang-ppm-dap-taskprov.md
cjpatton: Fair enough, but the following seems superfluous:

> In this case the leader MAY respond with HTTP status code 404
> Not Found and an error of type `unrecognizedTask`.

In vanilla DAP, if the task ID is not known, then the Leader is already required to abort with "unrecognizedTask". As far as I can tell, this text only changes the server behavior by allowing it to abort with 404 rather than the currently mandated 400.

> The response MAY include a
> Retry-After header field to suggest a pulling interval to the collector.

This seems fine, but a little odd. It seems like what you want to do here is distinguish between two cases:
1. A collect request is made for a task that will exist eventually (once we get the first upload request pertaining to the task)
2. A collect request is made for a task that will never exist

But how does the Leader know which of these is true, based on the information it has?

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,296 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+## Change to helper aggregate sub-protocol
 
+Upon receipt of a `AggregateInitializeReq`, helper reads the extension
+codepoint in `extension_type`. If helper does not support this extension, it
+MUST abort the aggregate protocol and alert the leader with error
+"unrecognizedMessage".
+
+> OPEN ISSUE: Alternatively Helpr MUST ignore the unrecognized extension, see
+> [#334](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/334) for
+> discussion.
```
cjpatton, line 323: Fine to point to #334, but I think for this PR we should align the Helper's behavior with the Leader's. In particular we should say that it MUST ignore the extension.

draft-wang-ppm-dap-taskprov.md
wangshan: It is defined in the Conventions and Definitions section:
> : The entity that defines the parameters of a task.

cjpatton: Sure, but we definitely need some "flavor text" to gently introduce the reader to this concept. We can worry about this later though :)

draft-wang-ppm-dap-taskprov.md
wangshan: that's a good point, we should probably aim to have collector behave the same in both cases. The core protocol today doesn't describe how leader response to a nonexistent task ID. The only `unrecognizedTask` error happens in hpke_config/ endpoint. I think the ideal solution could be the following:

- if leader receive a nonexistent task ID, it must responses 404 and `unrecognizedTask`, it may include Retry-After header.
- if collector gets 404 and `unrecognizedTask`, then it's up to collector to retry at a later time.

If this is specified in the core protocol, we can get rid of this collector section. thoughts? @cjpatton

cjpatton: I agree, it seems worth spelling this out in the core protocol. Would you mind sending a PR today? If not, I will do so.

One nit: Unless there's a specific reason why 404 is desirable, let's just go with 400, which is currently used for all aborts in DAP.

draft-wang-ppm-dap-taskprov.md
cjpatton: If the Helper doesn't recognize the task ID, it's supposed to abort with "unrecognizedTask". Are you sure this isn't already spelled out?

draft-wang-ppm-dap-taskprov.md
wangshan: I don't have a reason to use 404 but the current core protocol actually says it explicitly: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/main/draft-ietf-ppm-dap.md#hpke-configuration-request-hpke-config

> If the aggregator does not recognize the task ID, then it responds with HTTP status code 404 Not Found and an error of type unrecognizedTask.

wangshan: In the Errors table in the core protocol, it has this:
> unrecognizedTask | An endpoint received a message with an unknown task ID.

Is this considered "spelled out" (It doesn't actually say what HTTP status code to use)? If so, then there's nothing to be done.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,291 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+} VdafConfig;
+~~~
+
+> OPEN ISSUE: Should DP parameters be defined as a different "dimension" to
+> VDAF, given that it's likely various DP mechanisms can be applied to any VDAF.
+> See issue [#94](https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/94) for
+> discussion.
+
+The definition of Time, Duration, Url, QueryType follow those in
+{{?DAP=I-D.draft-ietf-ppm-dap-01}}.
```
junyechen1996, line 216:
```suggestion
{{?DAP=I-D.draft-ietf-ppm-dap-02}}.
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -56,76 +60,291 @@ collector). Thus it is up to each deployment of DAP to securely implement its
...
+        case fixed-size:     uint32 max_batch_size;
+    }
+} QueryConfig;
 ~~~
 
+The purpose of `TaskConfig` is to define all parameters that are necessary
+for configuring an aggregator. It includes all the fields to be
+associated with a task (see task configuration in
+{{?DAP=I-D.draft-ietf-ppm-dap-01}}.). Besides, `TaskConfig` also includes fields
+that useful for configuring a task in-band:
```
junyechen1996, line 172:
```suggestion
that are useful for configuring a task in-band:
```

junyechen1996 (approved):

draft-wang-ppm-dap-taskprov.md
cjpatton: This draft isn't out yet, so we'll have to update this later :)

junyechen1996: Fair enough, I'm just not clear because there is a mixture of reference to DAP-01 and DAP-02.

cjpatton: TOTALLY understandable! We were hoping to cut DAP-02 last week, but it looks like we'll merge this PR before then.

PR #7, opened 2022-09-14
Clean up TaskConfig definition

cjpatton:
- Wrie `Dp` (short for diff. privacy) instead of `Privacy`. That way we don't end up conflating this (precise) definition of privacy with other definitions. (For example, in the DAP draft, "privacy" refers to an MPC-style definition.)

- Write `Config` instead of `Params`, for consitency with `TaskConfig`, `QueryConfig`, and `VdafConfig`.

- Lift `DpConfig` out of `VdafConfig` as these apply to two separable primitives that we are composing in this application.

- Various editorial nits.

### DIFF
--- a/draft-wang-ppm-dap-taskprov.md
+++ b/draft-wang-ppm-dap-taskprov.md
@@ -122,7 +122,7 @@ enum {
 > See [#314](ietf-wg-ppm/draft-ietf-ppm-dap#314) for discussion.
 
 When the Client includes this extension with its report, the body of the
-extension is structured as follows:
+extension is a `TaskConfig` defined follows:
 
 ~~~
 struct {
@@ -148,9 +148,22 @@ struct {
     Time task_expiration;
 
     /* Determines the VDAF type and its config parameters. */
-    VdafConfig vdaf_config<1..2^16-1>;
+    VdafConfig vdaf_config;
 } TaskConfig;
+~~~
 
+The purpose of `TaskConfig` is to define all parameters that are necessary for
+configuring an aggregator. It includes all the fields to be associated with a
+task. (See task configuration in {{?DAP=I-D.draft-ietf-ppm-dap-01}}.) In
+addition to the aggregator endpoints, maximum batch lifetime, and task
+expriation, the structure includes an opaque `task_info` field that is specific
+to a deployment. For example, this can be a string describing the purpose of
+this task.
+
+The `query_config` field defines the DAP query configuration used to guide batch
+selection. It is defined as follows:
+
+~~~
 struct {
     QueryType query_type;    /* Defined in I-D.draft-ietf-ppm-dap-02 */
     Duration time_precision; /* Defined in I-D.draft-ietf-ppm-dap-02 */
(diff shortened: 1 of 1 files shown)


draft-wang-ppm-dap-taskprov.md
```diff
@@ -162,40 +179,19 @@ struct {
...
 
 ~~~
-/* Codepoint for each standardized VDAF. Defined in
-I-D.draft-irtf-cfrg-vdaf-03. */
 enum {
-    prio3-aes128-count(0x00000000),
-    prio3-aes128-sum(0x00000001),
-    prio3-aes128-histogram(0x00000002),
-    poplar1-aes128(0x00001000),
+    prio3-aes128-count(0x00000000),     // I-D.draft-irtf-cfrg-vdaf-03
```
wangshan, line 187: nit: why prefer inlined comment?

draft-wang-ppm-dap-taskprov.md
```diff
@@ -204,31 +200,32 @@ struct {
...
-when combined with privacy related parameters in `TaskConfig` like
-`min_batch_size`, can determine the VDAF's privacy guarantee.
-
 > OPEN ISSUE: Should spell out definition of `PrivacyParams` for various
 > differential privacy mechanisms and parameters.
 > See issue [#94](https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/94) for
 > discussion.
 
-The definition of Time, Duration, Url, QueryType follow those in
+The definition of `Time`, `Duration`, `Url`, QueryType follow those in
```
wangshan, line 228: you missed `QueryType`

wangshan (commented):
Left some minor comments, mostly about missed renaming & reformatting.

@cjpatton Why bringing DpConfig to TaskConfig, given that the exact mechanism is likely to be tied to a VDAF? DpConfig is not in the core protocol's out-of-band task configuration, if people wants to introduce a new Dp config and use it out-of-band, then it'll have to be aligned with the struct defined here.

draft-wang-ppm-dap-taskprov.md
cjpatton: It's easier to read, I think.

cjpatton:
> @cjpatton Why bringing DpConfig to TaskConfig, given that the exact mechanism is likely to be tied to a VDAF?

Fair point. I guess I was a bit confused about the requirements: If the expectation is that how we use DP is *always* VDAF-dependent, then this is a fair move; but there might be ways of applying central DP that apply to a large class of VDAFs (see https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/94). 

I've merged this back into vdaf_config. If at some point we want to lift this to task_config, we can do so in a way that's backwards compatible, since dp_config comprises the first bytes of vdaf_config.

> DpConfig is not in the core protocol's out-of-band task configuration, if people wants to introduce a new Dp config and use it out-of-band, then it'll have to be aligned with the struct defined here.

Good catch!

wangshan (approved):

PR #9, opened 2022-09-21
Flesh out some of the protocol details

cjpatton:
- Clean up the introduction. Be sure to mention all the salient points, including that some artifacts are still negotiated out of batnd.

- Spell out "opt-in" versus "opt-out" conditions. This makes the previous {{provisioning-a-task}} section more explicit and adds some additional features.

- Formalize the language in the protocol behavior sections.

- Add notes and an attack to security considerations.

- Move max_bach_lifetime into QueryConfig, as this parameter constrains queries.

- Various editorial things, including aligning with editorial changes in DAP-02.

### DIFF
--- a/draft-wang-ppm-dap-taskprov.md
+++ b/draft-wang-ppm-dap-taskprov.md
@@ -47,79 +47,85 @@ allows the task configuration to be provisioned in-band.
 
 # Introduction
 
-The DAP protocol {{?DAP=I-D.draft-ietf-ppm-dap-01}} enables secure aggregation
-of a set of reports submitted by clients. This process is centered around a
+The DAP protocol {{!DAP=I-D.draft-ietf-ppm-dap-01}} enables secure aggregation
+of a set of reports submitted by Clients. This process is centered around a
 "task" that determines, among other things, the cryptographic scheme to use for
 the secure computation (a Verifiable Distributed Aggregation Function
 {{?VDAF=I-D.draft-irtf-cfrg-vdaf-01}}), how reports are partitioned into
-batches, and privacy parameters such as the minumum size of each batch.
+batches, and privacy parameters such as the minimum size of each batch. Before a
+task can be executed, it is necessary to first provision the Clients,
+Aggregators, and Collector with the task's configuration.
 
-The core DAP specifcation does not define a mechanism for provisioning task
-configurations to the various parties (i.e., the clients, aggregators, and the
-collector). Thus it is up to each deployment of DAP to securely implement its
-own mechanism.
+However, The core DAP specification does not define a mechanism for provisioning
+tasks. this document describes a mechanism designed to fill this gap. Its key
+feature is that task configuration is performed completely in-band. It relies
+solely on the upload channel and the metadata carried by reports themselves.
 
-This document describes a mechanism for configuring tasks that may be useful in
-many deployments. The goal of this mechanism is to define a task provision
-method that utilizes just the upload channel and the metadata in the `Report`
-itself.
+This method presumes the existence of a logical "task author" (written as
+"Author" hereafter) who is capable of pushing configurations to Clients. All
+parameters required by downstream entities (the Aggregators and Collector) are
+encoded in an extension field of the Client's report. There is no need for
+out-of-band task orchestration between Leader and Helpers, therefore making
+adoption of DAP easier.
 
-At a high level, this extension asks client to include all the task
-configuration parameters it received out-of-band from task author, in the
-extension field of the `Report` it uploads to aggregators. The aggregators will
-create a DAP task upon receiving `Report` with such extension (or `ReportShare`
-in the case of helper.)
+The extension is designed with the same security and privacy considerations of
+the core DAP protocol. The Author is not regarded as a trusted third party: It
+is incumbent on all protocol participants to verify the task configuration
+disseminated by the Author and opt-out if the parameters are deemed insufficient
+for privacy. In particular, adopters of this extensions should presume the
+Author is under the adversary's control. In fact, we expect in a real-world
+deployment that the Author may be implemented by one of the Aggregators or
+Collector.
 
-By defining this mechanism as an extension, we can guarantee that for any
-deployments implementing this extension, all information required for
-aggregating a report is included in the report itself. There is no need for
-out-of-band task orchestration between leader and helpers, therefore making
-adoption of DAP easier. Since aggregators must create tasks using the same
-parameters client sent, this extension prevents task authors from giving clients
-and aggregators different task parameters in order to achieve reduced privacy
-guarantee, without requiring an out-of-band mechanism.
+Finally, the DAP protocol requires configuring the entities with a variety of
+assets that are not task-specific, but are important for establishing
+Client-Aggregator, Collector-Aggregator, and Aggregator-Aggregator
+relationships. These include:
 
-This extension affects upload, aggregate and collect sub-protocols.
+* The Collector's HPKE {{?RFC9180}} configuration used by the Aggregators to
+  encrypt aggregate shares
+
+* Any assets required for authenticating HTTP requests
+
+This specification does not specify a mechanism for provisioning these assets;
+as in the core DAP protocol, these are presumed to be configured out-of-band.
+
+Note that we consider the VDAF verification key {{!VDAF}}, used by the
+Aggregators to aggregate reports, to be a task-specific asset. This document
+specifies how to derive this key for a given task from a pre-shared secret,
+which in turn is presumed to be configured out-of-band.
 
 # Conventions and Definitions
 
 {::boilerplate bcp14-tagged}
 
-Similar to {{?DAP=I-D.draft-ietf-ppm-dap-01}}, this document uses the verbs
-"abort" and "alert with `[some error message]`" to describe how protocol
-participants react to various error conditions.
+This document uses the same conventions for error handling as
+{{!DAP=I-D.draft-ietf-ppm-dap-01}}. In addition, this document extends the core
+specification by adding the following error types:
 
-The terms used follow those described in {{?DAP=I-D.draft-ietf-ppm-dap-01}}. The
+| Type        | Description                                                                               |
+|:------------|:------------------------------------------------------------------------------------------|
+| invalidTask | An Aggregator has opted out of the indicated task as described in {{provisioning-a-task}} |
+
+The terms used follow those described in {{!DAP=I-D.draft-ietf-ppm-dap-01}}. The
 following new terms are used:
-
-Task provisioning:
-: The process of creating a DAP task.
 
 Task configuration:
 : The non-secret parameters required to create a task in task provision.
 
 Task author:
-: The entity that defines the parameters of a task.
+: The entity that defines a task's configuration.
 
-> TODO: clarify the definition of task author.
-
-# The "task-prov" Extension
-
-> NOTE(cjpatton) It will be useful to think of a good name for this. Something
-> succinct and descriptive is ideal. `task-prov` is fine, but we can maybe do
-> better.
+# The "task_prov" Extension
 
 A new extension is defined:
 
 ~~~
 enum {
-    task-prov(0xff00),
+    task_prov(0xff00),
     (65535)
 } ExtensionType;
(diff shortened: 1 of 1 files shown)


draft-wang-ppm-dap-taskprov.md
```diff
@@ -47,79 +47,85 @@ allows the task configuration to be provisioned in-band.
 
 # Introduction
 
-The DAP protocol {{?DAP=I-D.draft-ietf-ppm-dap-01}} enables secure aggregation
-of a set of reports submitted by clients. This process is centered around a
+The DAP protocol {{!DAP=I-D.draft-ietf-ppm-dap-01}} enables secure aggregation
```
wangshan, line 50: question: what's the difference between ! and ? in this format? is there a standard doc I can refer to for this kind of formatting?

cjpatton: "!" denotes "normative" reference; "?" denotes 'informative" reference. If a reference is "normative", then it is effectively part of the spec.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -222,144 +225,212 @@ struct {
...
+> so that we can treat the `task_id` as pseudorandom in cases where
+> `task_config` has high min-entropy
 
-To offer the `task-prov` extension, the client adds the `TaskConfig` structure
-it received from the task author in the extensions field of its `Report`. It
-computes the task ID as described in {{construct-task-id}}.
+> TODO: Add RFC reference for SHA-256.
 
-## Construct task ID {#construct-task-id}
+## Deriving the VDAF Verification Key {#vdaf-verify-key}
```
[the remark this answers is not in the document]
cjpatton: This is a question about the security of the VDAF. We are working on a paper that will address this question; I will keep https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/161 up-to-date.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -222,144 +225,212 @@ struct {
...
-If the decoding succeeds, aggregator creates a new task using the task ID
-from the decoded extension, and save task configuration with the newly created
-task. In particular, aggregator should deserialize `vdaf_config` corresponding
-to `vdaf_type`, and pass the relevant parameters to the VDAF initializer. At
-this point, the task provision step has completed.
+1. A secure connection to one or both of the Aggregator endpoints could not be
+   established.
 
-# Leader Behavior
+1. The task lifetime is too long.
```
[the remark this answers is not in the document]
cjpatton: The task_expiration is what I'm referring to. The behavior described here is: "A party MAY opt out if it thinks the task lives too long".

draft-wang-ppm-dap-taskprov.md
```diff
@@ -222,144 +225,212 @@ struct {
...
+
+The Collector might issue a collect request for a task provisioned by the
+`task_prov` extension prior to opting in to the task. In this case, the Leader
+would need to abort the collect request with "unrecognizedTask". When it does
+so, it SHOULD also include a "Retry-After" header in its HTTP response
+indicating the time after which the Collector should retry its request.
+
+> TODO: Find RFC reference for "Retry-After".
+
+> OPEN ISSUE: This semantics is awkward, as there's no way for the Leader to
```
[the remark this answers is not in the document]
cjpatton: Which text? The intent is that each party should opt in or out, including the Collector.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -222,144 +225,212 @@ struct {
...
-# Acknowledgments
+# Contributors
 {:numbered="false"}
 
-> NOTE(cjpatton) It's a good idea to acknowledge anyone by name who contributed
-> to the spec in some way, either directly or indirectly.
-> NOTE(shan) Will add more in future commits.
-
-Contributors
+> CP: Unless the order is meaningful, consider alphabetizing these names.
```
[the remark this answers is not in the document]
cjpatton: Raised issue

draft-wang-ppm-dap-taskprov.md
```diff
@@ -222,144 +225,212 @@ struct {
...
+"unrecognizedTask". Otherwise, if the Helper supports the extension, it proceeds
+as follows.
+
+First, the Helper checks that all report shares carried by the request pertain
+to the same task. In particular, it checks that:
+
+1. Either all report shares have the `task_prov` extension or none do. If not
+   the Helper MUST abort with "unrecognizedMessage".
+
+1. All report shares with the `task_prov` extension have the same payload. If
```
MichaelScaria, line 354: This should be "2. " right?

draft-wang-ppm-dap-taskprov.md
```diff
@@ -222,144 +225,212 @@ struct {
...
-## Change to helper aggregate sub-protocol
+> OPEN ISSUE: This awkward input validation step could be skipped if
+> `AggregateInitializeReq` had an extension field that we could stick the task
+> configuration in. This would also save significantly in overhead.
 
-Upon receipt of a `AggregateInitializeReq`, helper reads the extension
-codepoint in `extension_type`. If helper does not support this extension, it
-MUST ignore it. If the task ID is not known, then it MUST abort the aggregate
-protocol and alert the leader with error "unrecognizedTask".
+Next, the Helper attempts to parse the extension paylaod. If parsing fails, it
```
MichaelScaria, line 361: "payload" typo

draft-wang-ppm-dap-taskprov.md
cjpatton: `make` renders this as an enumerated list (1, 2, 3, ...). Using "1." to indicate an element of an enumerated list is better because you don't have to renumber the list when adding/removing items.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -222,144 +225,212 @@ struct {
...
 The "good" reports will not be polluted.
 
+> CP: I don't understand the threat here. Is it that the client may lie about
+> which task its report is for? The same is true for vanilla DAP, right?
+
+A malicious coalition of Clients might attempt to pollute an Aggregator's
+long-term storage by uploading reports for many (thousands or perhaps millions)
+of distinct tasks.
+
+> TODO: Suggest mitigations for this.
```
MichaelScaria, line 421: FWIW, we rate limit the number of "tasks" today.

MichaelScaria:
LGTM! Love to see this.

MichaelScaria (approved):

wangshan (changes requested):

draft-wang-ppm-dap-taskprov.md
```diff
@@ -222,144 +225,212 @@ struct {
...
+
+> OPEN ISSUE: What if the Collector and Aggregators don't agree on the task
+> configuration? Consider adding
 
 A "task" now means the same task ID and same task configuration, if a malicious
 client changes the task ID or task configuration, its report will be aggregated
 in a different task, with other poison reports from the same malicious attack.
 The "good" reports will not be polluted.
 
+> CP: I don't understand the threat here. Is it that the client may lie about
```
wangshan, line 414: Yes, in the core protocol, client has nothing to lie about task provision, since task has been provisioned before client knows about it. I think the main point is malicious client who wants to send fake extension data will never be mixed with clients that have legitimate task configurations.

draft-wang-ppm-dap-taskprov.md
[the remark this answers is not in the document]
cjpatton: My understanding is that task_expiration is basically a timestamp. If this is so, then I don't think this is ambiguous, since the MAY applies to expiration dates in the future, whereas the MUST applies to expiration dates in the past.

However there is another problem here. The goal of this MAY is to allow a party to opt-out if the task is going too long. There are two reasons why it might want to do this:
1. Sanity checking: We want to make sure tasks don't live for 100s of years.
2. For a given deployment of DAP, the parties might make an agreement that they will keep around metadata for anti-replay for as long as the task exists. In this case it would be useful to check that the task exists for, say, no longer than a week so that they don't have to keep around the metadata for longer than needed.

In either case, what's missing is the task's "start time". This is important because the Author might have created the task minutes or even hours before the Aggregators (in particular the Helper) sees the task for the first time.

If you agree this is needed, I'll add it to this PR.

draft-wang-ppm-dap-taskprov.md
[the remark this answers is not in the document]
cjpatton: If a non-taskprov Collector sends a CollectReq and gets unrecognizedTask in response, then it's just supposed to quit, since it's expectation is that there's no point in the future where the Leader will be configured with the task. Instead, we're asking it to retry at some point in the future, but retrying is necessarily fruitless. Since Retry-After is typically handled "under the hood" by HTTP clients (in the reqwest crate for Rust, you have to explicitly override this), non-taskprov Collectors will appear to hang indefinitely here.

This can be avoided if the Collector has to explicitly opt-in.

draft-wang-ppm-dap-taskprov.md
cjpatton: I see.  IMO this doesn't need to be called out in security considerations. In any case, we should try to clean this up.

draft-wang-ppm-dap-taskprov.md
cjpatton: Right, this TODO is asking us to spell out how to do this.

cjpatton: We can do this later, no need to do it in this PR.

draft-wang-ppm-dap-taskprov.md
wangshan: If it doesn't add anything to security considerations, I'm fine deleting this paragraph

draft-wang-ppm-dap-taskprov.md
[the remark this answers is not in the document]
wangshan: The leader can choose not to send retry-after header if it hasn't opt-in to task_prov, collector will just get unrecognizedTask and quit, the task_prov process happens inside leader/helper, so it should be sufficient for leader to decide whether a retry-after should be given to a collector.

draft-wang-ppm-dap-taskprov.md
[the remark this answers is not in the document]
wangshan: for case1, isn't the task_expiration sufficient? I'm not convinced a start time is the solution but I'm curious how this start time looks like, perhaps we can keep the current PR text and create another PR to address this?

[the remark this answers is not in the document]
cjpatton: > for case1, isn't the task_expiration sufficient?

Deciding whether a task has a lifetime of at least 200 years, I would subtract task_start from task_expiration and see if the duration is > 200 years. We can't do this correctly without knowing task_start.

> I'm not convinced a start time is the solution but I'm curious how this start time looks like, perhaps we can keep the current PR text and create another PR to address this?

Sure, makes sense! Raised https://github.com/wangshan/draft-wang-ppm-dap-taskprov/issues/12

draft-wang-ppm-dap-taskprov.md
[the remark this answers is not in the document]
cjpatton: How does the collector know that the task is definitely never going to exist? Remember, the intention of this retry-after mechanism is to allow the Leader to keep the Collector on the phone in case the task gets configured in the near future. If a taskprov Leader can't distinguish whether the Collector is taskprov or non-taskprov, it either has to ignore this SHOULD or risk keeping a non-taskprov Collector in an indefinite retry-after loop.

TBH, I would be happy to avoid this complexity by dropping the retry-after behavior altogether, at least until we can implement it unambiguously.

draft-wang-ppm-dap-taskprov.md
cjpatton: Ack, I'll leave as-is, we can clean up later.

cjpatton: https://github.com/wangshan/draft-wang-ppm-dap-taskprov/issues/13

wangshan (approved):

PR #17, opened 2022-09-29
Make select branch variable name unique

wangshan:
This maps well to any union based structs in implementation

### DIFF
--- a/draft-wang-ppm-dap-taskprov.md
+++ b/draft-wang-ppm-dap-taskprov.md
@@ -194,9 +194,9 @@ struct {
     VdafType vdaf_type;
     select (VdafConfig.vdaf_type) {
         case prio3_aes128_count: Empty;
-        case prio3_aes128_sum: uint8 bits;
-        case prio3_aes128_histogram: uint64 buckets<8..2^24-8>;
-        case poplar1_aes128: uint16 bits;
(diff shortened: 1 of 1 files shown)


draft-wang-ppm-dap-taskprov.md
```diff
@@ -194,9 +194,9 @@ struct {
     VdafType vdaf_type;
     select (VdafConfig.vdaf_type) {
         case prio3_aes128_count: Empty;
-        case prio3_aes128_sum: uint8 bits;
-        case prio3_aes128_histogram: uint64 buckets<8..2^24-8>;
-        case poplar1_aes128: uint16 bits;
+        case prio3_aes128_sum: uint8 prio3_aes128_sum_bits;
+        case prio3_aes128_histogram: uint64 prio3_aes128_histogram_buckets<8..2^24-8>;
+        case poplar1_aes128: uint16 poplar1_aes128_bits;
```
cjpatton, line 199: No objection, but I think this is too verbose. In fact, the following is valid (and preferable I think):
```suggestion
        case prio3_aes128_sum: uint8; /* bit length of the summand */
        case prio3_aes128_histogram: uint64<8..2^24-8>; /* bucket boundaries */
        case poplar1_aes128: uint16; /* bit length of input string */
```

cjpatton (approved):

PR #19, opened 2022-10-05
Use HKDF to extract task ID and VDAF verify key

### LINKED ISSUE #4: Flesh out VDAF verificaiton key derivation
Since we're using SHA2 for the task ID, we might as well use HKDF-SHA2-Expand for deriving the verification key.

cjpatton:
Closes #4.


wangshan (approved):
LGTM

Issue #11, opened 2022-09-26
Advise deployments to support a global HPKE config if in-band task provisioning is supported


junyechen1996:
If a deployment supports in-band task provisioning, then clients should not query `/hpke_config` with a specified `task_id`. The reason is:

1. Aggregators need the first client report for a particular TaskID, in order for the task to be configured.
2. Clients need the HPKE config to submit reports. If `task_id` parameter is included when clients query `/hpke_config` endpoint, the Aggregator may not know that TaskID yet, because it has not received any report for that.

One way is to advise deployment to support a global HPKE config, and also advise clients to not pass `task_id` to `/hpke_config` endpoint. The other way is to create the HPKE config on the fly for unseen task ID in `/hpke_config`.

cjpatton:
Yup 👍

cjpatton:
@junyechen1996 we just merged #9  so this would be a good time for text

junyechen1996:
Addressed by https://github.com/wangshan/draft-wang-ppm-dap-taskprov/pull/21

PR #21, opened 2022-10-06
Address global HPKE config - Issue #11

junyechen1996:


draft-wang-ppm-dap-taskprov.md
```diff
@@ -296,6 +296,23 @@ A protocol participant MUST opt out if the task has expired.
 The behavior of each protocol participant is determined by whether or not they
 opt in to a task.
 
+## Supporting a Global HPKE Configuration {#global-hpke-config}
+
+In DAP, Clients need to know the HPKE configuration of each Aggregator before
+sending reports (See HPKE Configuration Request in
```
cjpatton, line 302:
```suggestion
sending reports. (See HPKE Configuration Request in
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -296,6 +296,23 @@ A protocol participant MUST opt out if the task has expired.
...
+{{!DAP=I-D.draft-ietf-ppm-dap-02}}.)
+
+If `task_prov` extension is supported in a DAP deployment, each Aggregator
+SHOULD support a global HPKE configuration independent of Task ID, because
```
cjpatton, line 306:
```suggestion
SHOULD support a global HPKE configuration independent of the task ID, because
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -296,6 +296,23 @@ A protocol participant MUST opt out if the task has expired.
 The behavior of each protocol participant is determined by whether or not they
 opt in to a task.
 
+## Supporting a Global HPKE Configuration {#global-hpke-config}
+
```
cjpatton, line 300: editorial: I suggest reworking this so that all of the motivation is in the first paagragh, then all normative text follows:

> Clients need to know the HPKE configuration prior to uploading reports, but 
>
> To mitigate this issue, Aggregators SHOULD ...
>
> In addition, Clients SHOULD NOT ...

draft-wang-ppm-dap-taskprov.md
```diff
@@ -296,6 +296,23 @@ A protocol participant MUST opt out if the task has expired.
...
+In DAP, Clients need to know the HPKE configuration of each Aggregator before
+sending reports (See HPKE Configuration Request in
+{{!DAP=I-D.draft-ietf-ppm-dap-02}}.)
+
+If `task_prov` extension is supported in a DAP deployment, each Aggregator
+SHOULD support a global HPKE configuration independent of Task ID, because
+a task is only configured in an Aggregator after the Aggregator has received
+the first Client report that contains the `task_prov` extension.
+
+If a Client wants to query an Aggregator's `[aggregator]/hpke_config` endpoint,
```
cjpatton, line 310: There is no reason to repeat the endpoint format here. The more we can refer to the DAP draft as the source of truth, the better.

cjpatton (changes requested):
I'm not sure "SHOULD use a global config" is the right guidance. It's possible that a single Aggregator would support taskprov tasks as well as non-taskprov tasks. Indeed, an Aggregator might want to be able to enforce key separation between tasks that are provisioned by taskprov and those that are not. Perhaps we can try to be more specific about this guidance?

draft-wang-ppm-dap-taskprov.md
junyechen1996: Good point, there are three paragraphs: 1. describe the issue of specifying `task_id` in `/hpke_config`, 2. Aggregator's recommended behavior, 3. Client's recommended behavior.

junyechen1996:
> I'm not sure "SHOULD use a global config" is the right guidance. It's possible that a single Aggregator would support taskprov tasks as well as non-taskprov tasks. Indeed, an Aggregator might want to be able to enforce key separation between tasks that are provisioned by taskprov and those that are not. Perhaps we can try to be more specific about this guidance?

I added some text to recommend an Aggregator "MAY continue to support per-task HPKE configurations for other tasks that are configured out-of-band". Let me know if this makes things clearer.

> 
> Another more minor issue: An Aggregator might want to periodically rotate the HPKE config it advertises to taskprov Clients.

I think this can be up to the Aggregator, because key rotation needs to happen for per-task HPKE config as well. I can add some text to recommend key rotation though, let me know what you think.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -296,6 +296,27 @@ A protocol participant MUST opt out if the task has expired.
 The behavior of each protocol participant is determined by whether or not they
 opt in to a task.
 
+## Supporting a Global HPKE Configuration {#global-hpke-config}
+
+In DAP, Clients need to know the HPKE configuration of each Aggregator before
+sending reports. (See HPKE Configuration Request in
+{{!DAP=I-D.draft-ietf-ppm-dap-02}}.) However, in a DAP deployment that supports
+`task_prov` extension, if a Client queries an Aggregator's `/hpke_config`
+endpoint with a task ID computed as described in {{construct-task-id}}, the
```
cjpatton, line 305: It's best to avoid writing text that refers to specific constructions in the base protocol. Suppose for example that the core protocol changed the endpoint to `/get_that_cool_hpke_config`: Then we would have to update this reference in-kind. My suggestion avoids this.
```suggestion
`task_prov` extension, if a Client requests the Aggregator's HPKE configuration
with the task ID computed as described in {{construct-task-id}}, the
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -296,6 +296,27 @@ A protocol participant MUST opt out if the task has expired.
...
+
+To mitigate this issue, if an Aggregator wants to support `task_prov` extension,
+it SHOULD support a global HPKE configuration independent of the task ID, for all
+tasks that are configured with `task_prov` extension. It MAY continue to support
+per-task HPKE configurations for other tasks that are configured out-of-band.
+
+In addition, if a Client wants to include `task_prov` extension in its report,
+it SHOULD NOT specify the `task_id` parameter when querying an Aggregator's
+`/hpke_config` endpoint, because the task ID may not have been configured
+in the Aggregator yet.
```
cjpatton, line 318: I think the reason is clear at this point.

```suggestion
`/hpke_config` endpoint.
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -296,6 +296,27 @@ A protocol participant MUST opt out if the task has expired.
...
+sending reports. (See HPKE Configuration Request in
+{{!DAP=I-D.draft-ietf-ppm-dap-02}}.) However, in a DAP deployment that supports
+`task_prov` extension, if a Client queries an Aggregator's `/hpke_config`
+endpoint with a task ID computed as described in {{construct-task-id}}, the
+task ID may not be configured in the Aggregator yet, because the Aggregator is
+still waiting for the first Client report with the `task_prov` extension
+to arrive.
+
+To mitigate this issue, if an Aggregator wants to support `task_prov` extension,
+it SHOULD support a global HPKE configuration independent of the task ID, for all
```
cjpatton, line 311: Re: https://github.com/wangshan/draft-wang-ppm-dap-taskprov/pull/21#issuecomment-1271070699

The problem is the guidance "SHOULD support a global HPKE configuration". We don't want to limit the Aggregator to just one configuration, for the following reasons:
1. It may want to support multiple HPKE cipher suites (say, X25519 and P256)
2. It may want to support multiple HPKE keys at the same time (say, to account for clients that may have used an older config)

Something like this might be better: "SHOULD choose which HPKE configuration to advertise to clients independent of the task ID."

draft-wang-ppm-dap-taskprov.md
junyechen1996: Also changed `/hpke_config` to "when requesting the HPKE configuration from an Aggregator".

draft-wang-ppm-dap-taskprov.md
junyechen1996: I see, I didn't realize that is not clear. Maybe we should avoid the phrase "a global HPKE configuration".

I applied your suggestion, and also updated the title of this section to `Supporting HPKE Configurations Independent of Tasks {#hpke-config-no-task-id}`. Let me know if that doesn't make sense.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -296,6 +296,26 @@ A protocol participant MUST opt out if the task has expired.
 The behavior of each protocol participant is determined by whether or not they
 opt in to a task.
 
+## Supporting HPKE Configurations Independent of Tasks {#hpke-config-no-task-id}
+
+In DAP, Clients need to know the HPKE configuration of each Aggregator before
+sending reports. (See HPKE Configuration Request in
+{{!DAP=I-D.draft-ietf-ppm-dap-02}}.) However, in a DAP deployment that supports
+`task_prov` extension, if a Client requests the Aggregator's HPKE configuration
```
cjpatton, line 304:
```suggestion
the `task_prov` extension, if a Client requests the Aggregator's HPKE configuration
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -296,6 +296,26 @@ A protocol participant MUST opt out if the task has expired.
...
+task ID may not be configured in the Aggregator yet, because the Aggregator is
+still waiting for the first Client report with the `task_prov` extension
+to arrive.
+
+To mitigate this issue, if an Aggregator wants to support `task_prov` extension,
+it SHOULD choose which HPKE configuration to advertise to Clients independent of
+the task ID. It MAY continue to support per-task HPKE configurations for other
+tasks that are configured out-of-band.
+
+In addition, if a Client wants to include `task_prov` extension in its report,
```
cjpatton, line 315:
```suggestion
In addition, if a Client wants to include the `task_prov` extension in its report,
```

cjpatton (commented):
Looks good, just a couple more nits.

cjpatton (approved):
I'd suggest squashing all of the commits into one before merging.

junyechen1996:
I just squashed all commits into one.

PR #26, opened 2022-10-22
Use "dap-taskprov" for the salt constant

cjpatton:
We use "dap-taskprov-00" right now, the idea being that we would revise this to "dap-taskprov-01" for the next draft. Shan suggests this might end up being a footgun when we have to update to the next draft.

### DIFF
--- a/draft-wang-ppm-dap-taskprov.md
+++ b/draft-wang-ppm-dap-taskprov.md
@@ -237,7 +237,7 @@ task_id = HKDF-Extract(task_prov_salt, task_config)
 
 where `task_config` is the `TaskConfig` structure disseminated by the Author.
 String `task_prov_salt` is defined to be the SHA-256 hash of the octet string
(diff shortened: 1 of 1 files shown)


draft-wang-ppm-dap-taskprov.md
```diff
@@ -237,7 +237,7 @@ task_id = HKDF-Extract(task_prov_salt, task_config)
 
 where `task_config` is the `TaskConfig` structure disseminated by the Author.
 String `task_prov_salt` is defined to be the SHA-256 hash of the octet string
-"dap-taskprov-00". Function HKDF-Extract() denotes the HKDF extraction function
+"dap-taskprov". Function HKDF-Extract() denotes the HKDF extraction function
```
cjpatton, line 240: I have no objection to this change. If we need a hard-break between versions, then we can always update the extension codepoint.

wangshan (approved):

PR #30, opened 2023-03-09
Align taskprov with latest DAP draft

cjpatton:
* Bump references to draft-ietf-ppm-dap-02 to 04

* Bump references to draft-irtf-cfrg-vdaf-03 to 05

* The latest DAP draft drops metadata extensions in favor of per-Aggregator extensions included in the report shares. These new extensions are not well-suited for shared extensions, since we end up blowing up communication cost by transmitting the task configuration twice in the same report.

  Including the task configuration in the extensions has another problem: All reports include it, which means it is necessary for the Helper to check that each report in an aggregation job includes the exact same copy of this extension. This adds unnecessary complexity to the spec.

  We solve these problems by advertising the task configuration in a header: The Client includes it in its upload request to he Helper, and the Leader includes it in the first aggregate request incident to the task.

* Remove a security consideration about HKDF that no longer applies since we're using SHA-256 for task ID derivation.

* Minor editorial nits.

### DIFF
--- a/draft-wang-ppm-dap-taskprov.md
+++ b/draft-wang-ppm-dap-taskprov.md
@@ -53,19 +53,19 @@ allows the task configuration to be provisioned in-band.
 
 # Introduction
 
-The DAP protocol {{!DAP=I-D.draft-ietf-ppm-dap-02}} enables secure aggregation
+The DAP protocol {{!DAP=I-D.draft-ietf-ppm-dap-04}} enables secure aggregation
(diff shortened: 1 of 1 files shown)


cjpatton:
Note: Tests will fail until we have cut VDAF-05 and DAP-04.

wangshan (approved):

cjpatton:
Pushed a couple of formatting issues.

PR #38, opened 2023-09-05
Align with draft-ietf-ppm-dap-06

### LINKED ISSUE #33: New histogram format in VDAF-06
The latest VDAF draft (version 6) modified the format for Prio3Histogram: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-06#name-prio3histogram. To create a new Prio3Histogram we only need to know the number of buckets, and we no longer keep track of a vector of bucket boundaries. Measurements are now the bucket index (e.g. here is Libprio's implementation: https://github.com/divviup/libprio-rs/blob/a52c7d56e05741ae852166fd2cf2c223bc0c732b/src/vdaf/prio3_test.rs#L53). 

This no longer matches TaskProv's `VdafConfig.vdaf_type`: https://wangshan.github.io/draft-wang-ppm-dap-taskprov/draft-wang-ppm-dap-taskprov.html#section-3-9.

### LINKED ISSUE #36: Update histogram representation for VDAF-06
Taskprov currently has:
```
struct {
    DpConfig dp_config;
    VdafType vdaf_type;
        case prio3_aes128_histogram: uint64<8..2^24-8>; /* buckets */
    }
} VdafConfig;
```

In VDAF-06, rather than taking a list of bucket boundaries, the `Prio3Histogram` VDAF just takes a length, which is the number of buckets.

Additionally, all these VDAF names have `aes128` in them, which is also obsolete.

### LINKED ISSUE #37: Future changes to VDAF algorithm IDs
I wanted to give you a heads up that we're planning on changing the algorithm ID of Prio3Histogram in a future VDAF draft as part of cfrg/draft-irtf-cfrg-vdaf#284. This will affect the `VdafType` enum in future versions of this document.

cjpatton:
Closes #37.
Closes #33.

### DIFF
--- a/draft-wang-ppm-dap-taskprov.md
+++ b/draft-wang-ppm-dap-taskprov.md
@@ -53,11 +53,11 @@ allows the task configuration to be provisioned in-band.
 
 # Introduction
 
-The DAP protocol {{!DAP=I-D.draft-ietf-ppm-dap-04}} enables secure aggregation
+The DAP protocol {{!DAP=I-D.draft-ietf-ppm-dap-06}} enables secure aggregation
 of a set of reports submitted by Clients. This process is centered around a
 "task" that determines, among other things, the cryptographic scheme to use for
 the secure computation (a Verifiable Distributed Aggregation Function
-{{!VDAF=I-D.draft-irtf-cfrg-vdaf-05}}), how reports are partitioned into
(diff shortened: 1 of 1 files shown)


wangshan (approved):
Looks good to me!

PR #45, opened 2023-10-18
Opt-out if a task parameter is not recognized

### LINKED ISSUE #23: Decoding unrecognized message variants
The following messages have contents for which decoding depends on an `enum`:
- VdafConfig (need to know the VdafType in order to decode the rest of the message)
- DpConfig (need to know the DpMechanism in order to decode the rest of the message)
- QueryConfig (need to know the QueryType in order to decode everything after min_batch_size)

What do we do if we get a variant we don't recognize? How do we decode the rest of the message? The answer is we can't, since we don't know how many more bytes to consume. All we can do in taskprov-00 is abort with "unrecgonizedMessage", since we don't know how to parse.

Question 1: Is this the intended behavior? Or would it be better "opt-out" by aborting with "invalidTask"?

Regardless of the answer for taskprov-00, I think for the next version (taskprov-01) we ought to make sure messages can always be decoded, regardless of whether we know how to handle the parameters. The easiest way to do this would be to add a length prefix so that the contents can be handled as opaque blobs. This might look something like this (I'll have to double check if this is valid syntax; a similar situation arises in the TLS 1.3 record layer):

```
struct {
       DpConfig dp_config;
       VdafType vdaf_type;
+      uint16_t vdaf_param_len;
           case prio3_aes128_histogram: uint64<8..2^24-8>; /* buckets */
+          default: opaque[VdafConfig.vdaf_param_len];
       }
   } VdafConfig;
```

This came up while reviewing changes for Daphne: https://github.com/cloudflare/daphne/pull/148#discussion_r1001182181.

cjpatton:
Closes #23.

Previously we would fail to parse the `TaskConfig` if one of its parameters was not recognized. Modify the wire format so that it is always decodable.

### DIFF
--- a/draft-wang-ppm-dap-taskprov.md
+++ b/draft-wang-ppm-dap-taskprov.md
@@ -183,9 +183,11 @@ struct {
     Duration time_precision;      /* I-D.draft-ietf-ppm-dap-07 */
     uint16 max_batch_query_count; /* I-D.draft-ietf-ppm-dap-07 */
     uint32 min_batch_size;
+    uint16_t query_type_param_len; /* length of the remainder */
     select (QueryConfig.query_type) {
         case time_interval: Empty;
         case fixed_size:    uint32 max_batch_size;
+        default:            opaque[QueryConfig.query_type_param_len];
     }
 } QueryConfig;
 ~~~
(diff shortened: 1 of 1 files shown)


wangshan (approved):
LGTM

PR #49, opened 2023-11-16
QueryConfig: Move query type to just before the params

cjpatton:
This is intended to simplify encoding/decoding of this struct.


wangshan (approved):

cjpatton:
Rebased.

PR #58, opened 2024-01-12
Change dp_config to opaque

junyechen1996:
Also remove length hints in TLS variants that are mandatory to configure a task.

### DIFF
--- a/draft-wang-ppm-dap-taskprov.md
+++ b/draft-wang-ppm-dap-taskprov.md
@@ -183,16 +183,12 @@ struct {
     uint16 max_batch_query_count; /* I-D.draft-ietf-ppm-dap-07 */
     uint32 min_batch_size;
     QueryType query_type;        /* I-D.draft-ietf-ppm-dap-07 */
-    uint16 query_type_param_len; /* length of the remainder */
     select (QueryConfig.query_type) {
         case time_interval: Empty;
         case fixed_size:    uint32 max_batch_size;
     };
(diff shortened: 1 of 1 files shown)


junyechen1996:
I'm mainly trying to align with our implementations, but I'd welcome feedback on the texts here.

wangshan (approved):

cjpatton:
I think this is a regression. The goal of encoding the length of query type, vdaf type, and dp type is so that we can decode a variant we don't recognize. The goal was to separate parsing from protocol logic that is implementation specific.

I think this is a regression.

Thanks @junyechen1996 for pointing this out.

cjpatton:
@wangshan you might remember we discussed this a few months ago.

wangshan:
@cjpatton this isn't a regression, what we had before was a DpConfig with no length so decoder may not know how to skip it. What we write here uses `opaque dp_config<1..2^16-1>` which encodes the length in the variant length array already. When decoding, you either know how to decode dp_config or know how long it is by reading the first 2 bytes.

cjpatton:
> @cjpatton this isn't a regression, what we had before was a DpConfig with no length so decoder may not know how to skip it. What we write here uses `opaque dp_config<1..2^16-1>` which encodes the length in the variant length array already. When decoding, you either know how to decode dp_config or know how long it is by reading the first 2 bytes.

The regression is: There is no length prefix for VdafConfig or QueryConfig. DpConfig is fine.

PR #60, opened 2024-01-17
Add back `DpConfig` and the length prefixes in `VdafConfig` and `QueryConfig`

cjpatton:
**Option 1.**

A previous commit f8cedf129c0dfb6a3461b33f90717fcf88cc534b, the `DpConfig` structure was replaced with an opaque byte string that the parties are meant to ignore. We don't want the parties to ignore unrecognized parameters: they should opt out instead. This change specifies that the content should be a `DpConfig`. We also remove the length prefix from that structure, since it's no longer needed.

The same commit removed the length prefixes from `VdafConfig` and `QueryConfig`. This is a regression because it means the encoding of the `TaskConfig` is no longer self-describing. In particular, it forces to conflate opt-out due to unrecognized variants (an unimplemented VDAF, say) with serialization errors. Solve this by adding back the length prefixes.

### DIFF
--- a/draft-wang-ppm-dap-taskprov.md
+++ b/draft-wang-ppm-dap-taskprov.md
@@ -183,12 +183,16 @@ struct {
     uint16 max_batch_query_count;
     uint32 min_batch_size;
     QueryType query_type;
+    uint16 query_type_param_len; /* length of the remainder */
     select (QueryConfig.query_type) {
         case time_interval: Empty;
         case fixed_size:    uint32 max_batch_size;
     };
 } QueryConfig;
 ~~~
+
+The value `query_type_param_len` field MUST match length of the remainder of
+the structure.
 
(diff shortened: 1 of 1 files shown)


draft-wang-ppm-dap-taskprov.md
```diff
@@ -213,6 +217,7 @@ enum {
 struct {
     opaque dp_config<1..2^16-1>;  /* Encoded differential privacy parameters */
```
cjpatton, line 218: Reviewer note: it would be slightly more consistent to replace this with `DpConfig` and add the internal length prefix back (it was there before https://github.com/wangshan/draft-wang-ppm-dap-taskprov/commit/f8cedf129c0dfb6a3461b33f90717fcf88cc534b), but I wanted to the minimal change here.

PR #66, opened 2024-06-28
Issue 48: Acknowledge security consideration of having task Author

### LINKED ISSUE #48: Privacy considerations: Tracking users across tasks
@bemasc pointed out the following attack at IETF 118. A malicious Author can provision a Client with a unique task that allows it to track uploads from the Client over time. This might allow it to collude with the Leader and track where they are.

Is this an attack we're concerned about?

junyechen1996:
Closes #48.

### DIFF
--- a/draft-wang-ppm-dap-taskprov.md
+++ b/draft-wang-ppm-dap-taskprov.md
@@ -579,13 +579,26 @@ Author to be under control of the adversary. It is therefore incumbent on
 protocol participants to verify the privacy parameters of a task before opting
 in.
 
-Another risk is that a malicious coalition of Clients might attempt to pollute
-an Aggregator's long-term storage by uploading reports for many (thousands or
-perhaps millions) of distinct tasks. While this does not directly impact tasks
-used by honest Clients, it does present a Denial-of-Service risk for the
-Aggregators themselves. This can be mitigated by limiting the rate at which new
-tasks or configured. In addition, deployments SHOULD arrange for the Author to
-digitally sign the task configuration so that Clients cannot forge task creation.
+Another risk is that the Author could configure a unique task to fingerprint a
+Client. Although Client anonymization is not guaranteed by DAP, some systems
+built on top of DAP may hope to achieve this property by using a proxy server
+with Oblivious HTTP {{!RFC9458}} to forward Client reports to the Leader. If the
+Author colludes with the Leader, the attacker can learn some metadata
+information about the Client, e.g., the Client IP, user agent string, which may
+deanonymize the Client. However, even if the Author succeeds in doing so, the
+Author should learn nothing other than the fact that the Client has uploaded a
+report, assuming the Client has verified the privacy parameters of the task
+before opting into it. For example, if a task is uniquely configured for the
+Client, the Client can enforce the minimum batch size is strictly more than 1.
+
+Another risk for the Aggregators is that a malicious coalition of Clients might
+attempt to pollute an Aggregator's long-term storage by uploading reports for
+many (thousands or perhaps millions) of distinct tasks. While this does not
+directly impact tasks used by honest Clients, it does present a
+Denial-of-Service risk for the Aggregators themselves. This can be mitigated by
+limiting the rate at which new tasks or configured. In addition, deployments
+SHOULD arrange for the Author to digitally sign the task configuration so that
+Clients cannot forge task creation.
 
 # Operational Considerations
 


draft-wang-ppm-dap-taskprov.md
```diff
@@ -568,7 +568,10 @@ by including a new logical role, called the Author. The Author is responsible
 for configuring Clients prior to task execution. For privacy we consider the
 Author to be under control of the adversary. It is therefore incumbent on
 protocol participants to verify the privacy parameters of a task before opting
-in.
+in. One risk introduced by the addition of the Author is the Author can
+configure tasks uniquely for a Client. If the Author colludes with the Leader,
+it can track the uploading activities of that Client, which leaks information
+about that Client, e.g., location.
```
junyechen1996, line 574: We can point to some mitigation, e.g., OHTTP, not sure if it's worth mentioning here.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -568,7 +568,10 @@ by including a new logical role, called the Author. The Author is responsible
 for configuring Clients prior to task execution. For privacy we consider the
 Author to be under control of the adversary. It is therefore incumbent on
 protocol participants to verify the privacy parameters of a task before opting
```
cjpatton, line 570: Let's make this a new paragraph.

draft-wang-ppm-dap-taskprov.md
cjpatton: How does OHTTP mitigate this risk?

junyechen1996: Actually, by checking the task parameters, it will be mitigated slightly, e.g., by making sure min batch size is not too small, but the Author can still target the same group of clients for this attack.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -570,13 +570,19 @@ Author to be under control of the adversary. It is therefore incumbent on
...
 in.
 
-Another risk is that a malicious coalition of Clients might attempt to pollute
-an Aggregator's long-term storage by uploading reports for many (thousands or
-perhaps millions) of distinct tasks. While this does not directly impact tasks
-used by honest Clients, it does present a Denial-of-Service risk for the
-Aggregators themselves. This can be mitigated by limiting the rate at which new
-tasks or configured. In addition, deployments SHOULD arrange for the Author to
-digitally sign the task configuration so that Clients cannot forge task creation.
+One risk introduced by the addition of the Author is the Author can configure
```
cjpatton, line 573:
```suggestion
Another risk introduced by the addition of the Author is the Author could configure
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -570,13 +570,19 @@ Author to be under control of the adversary. It is therefore incumbent on
...
+tasks uniquely for a Client. If the Author colludes with the Leader, it can
+track the uploading activities of that Client, which leaks information about
+that Client, e.g., location.
```
cjpatton, line 576:
```suggestion
track the uploading activities of that Client, which could leak information about
that Client, e.g., location.
```

cjpatton, line 576:
```suggestion
track the uploading activities of that Client, which leaks information about
that Client, e.g., its location.
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -570,13 +570,19 @@ Author to be under control of the adversary. It is therefore incumbent on
 protocol participants to verify the privacy parameters of a task before opting
 in.
```
cjpatton, line 572: It's not clear to me why tracking a client across uploads leaks the client's location. Do you have an idea in mind? 

I realize this is mentioned in #48, but now I'm not sure it's the best example. Stepping back, what does an attacker that controls the Author (and possible the Collector and an Aggregator) by learn by fingerprinting a Client's upload task? Perhaps this would violate other properties adjcent to DAP, like anonymity or differential privacy?

draft-wang-ppm-dap-taskprov.md
cjpatton: (moved to another thread)

draft-wang-ppm-dap-taskprov.md
junyechen1996: I think if a client sends report to leader directly, then the upload will contain things like IP, or user agent, etc. That can leak the client location, device. And by not checking task parameters carefully, as a task author, I can set min_batch_size to be 1, and see the client measurement directly, along with user agent and IP.

cjpatton: > I think if a client sends report to leader directly, then the upload will contain things like IP, or user agent, etc.

Right: DAP on its own doesn't provide for anonymity. Suppose though you rand DAP on top of OHTTP.

> That can leak the client location, device. And by not checking task parameters carefully, as a task author, I can set min_batch_size to be 1, and see the client measurement directly, along with user agent and IP.

This is supposed to be prevented by the "opt-in" procedure.

junyechen1996: I guess maybe looking at this issue again, do we just care about the capability of author colluding with leader, where author is not a collector? So it cannot see the measurement in the clear. So you can only get the metadata information of a particular client, e.g. IP and user agent. Is that the only security risk we care about then?

cjpatton: Yeah I think that's a good example of the kind of risk we're thinking of. More to the point: The metadata may convey to the Leader who the client is; and if you're expecting that to be impossible, say because you use OHTTP, then you have to be very careful about how taskprov is deployed.

cjpatton: But I think we could also make the point that, even if the attacker can deanonymize a client, it's hard for it do learn anything more than the client uploaded a report. If the task is unique to that client, and the batch size is larger than 1, then it's really not going to get any more information.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -570,13 +570,26 @@ Author to be under control of the adversary. It is therefore incumbent on
...
-used by honest Clients, it does present a Denial-of-Service risk for the
-Aggregators themselves. This can be mitigated by limiting the rate at which new
-tasks or configured. In addition, deployments SHOULD arrange for the Author to
-digitally sign the task configuration so that Clients cannot forge task creation.
+Another risk introduced by the addition of the Author is the Author could
+configure tasks uniquely for a Client to fingerprint the Client. If the Author
+colludes with the Leader, it can learn some metadata information about the
+Client, e.g., the Client IP and user agent string. This can be mitigated by
+anonymizing the Client, such as by using Oblivious HTTP
+{{!I-D.draft-ietf-ohai-ohttp-08}} in the DAP upload route. We also note that,
```
cjpatton, line 578: It's not clear to me that OHTTP mitigates this issue. In fact, I think the converse is true: fingerprinting the client effectively removes any protection that the client gets from OHTTP. Concretely, the Author configures the client with `TaskConfig<task_info = "this client has IP address 192.168.0.1">`; when the Leader gets the Client's report (via the OHTTP proxy), the Leader immediately knows the IP address of the client that produced the report.

junyechen1996: You mean only targeting that client in a particular IP address. Let's avoid talking about OHTTP mitigation in this scenario then. Instead we will just say, if the author can deanonymize the client, the author can learn the metadata information at best, if the client carefully checks for privacy parameter.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -579,13 +579,24 @@ Author to be under control of the adversary. It is therefore incumbent on
...
-Another risk is that a malicious coalition of Clients might attempt to pollute
-an Aggregator's long-term storage by uploading reports for many (thousands or
-perhaps millions) of distinct tasks. While this does not directly impact tasks
-used by honest Clients, it does present a Denial-of-Service risk for the
-Aggregators themselves. This can be mitigated by limiting the rate at which new
-tasks or configured. In addition, deployments SHOULD arrange for the Author to
-digitally sign the task configuration so that Clients cannot forge task creation.
+Another risk is that the Author could configure a unique task to fingerprint a
+Client. If the Author colludes with the Leader, it can learn some metadata
+information about the Client, e.g., the Client IP, user agent string, or the
```
cjpatton, line 584: Is this what you mean?
```suggestion
Client. If the Author colludes with the Leader, the Leader can learn some metadata
information about the Client, e.g., the Client IP, user agent string, or the
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -579,13 +579,24 @@ Author to be under control of the adversary. It is therefore incumbent on
...
+upload timestamp. However, even though the Author is able to deanonymize the
```
cjpatton, line 585: The leader already learns the upload timestamp. (It's in the report metadata.)

Also, let's be more precise about how it learns things it's not supposed to.

cjpatton (commented):
This is looking good. One question we need to answer: under what circumstances is this attack a problem? Anonymity of the clients is not one of DAP's core design goals. On the other hand, when we compose DAP with something like OHTTP, this attack threatens the anonymity properties OHTTP is supposed to provvide.

draft-wang-ppm-dap-taskprov.md
junyechen1996: You mean "the Author"?

cjpatton: Actually I meant the Leader, but Author makes sense too I suppose. Perhaps we should just say "attacker" here.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -579,13 +579,27 @@ Author to be under control of the adversary. It is therefore incumbent on
...
-an Aggregator's long-term storage by uploading reports for many (thousands or
-perhaps millions) of distinct tasks. While this does not directly impact tasks
-used by honest Clients, it does present a Denial-of-Service risk for the
-Aggregators themselves. This can be mitigated by limiting the rate at which new
-tasks or configured. In addition, deployments SHOULD arrange for the Author to
-digitally sign the task configuration so that Clients cannot forge task creation.
+Another risk is that the Author could configure a unique task to fingerprint a
+Client. Although Client anonymization is not guaranteed by DAP, some systems
+built on top of DAP may hope to achieve this property by using a proxy server
+with Oblivious HTTP {{!I-D.draft-ietf-ohai-ohttp-08}} to forward Client reports
```
cjpatton, line 585: :)

```suggestion
with Oblivious HTTP {{!RFC9458}} to forward Client reports
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -579,13 +579,27 @@ Author to be under control of the adversary. It is therefore incumbent on
...
+to the Leader. If the Author colludes with the Leader, the Author can learn some
```
cjpatton, line 586:
```suggestion
to the Leader. If the Author colludes with the Leader, the attacker can learn some
```

cjpatton (approved):
Thanks for being patient with me!

draft-wang-ppm-dap-taskprov.md
junyechen1996: I copied it from DAP draft, perhaps this needs to be updated in DAP as well.

junyechen1996:
Squashed.

PR #67, opened 2024-06-28
Issue 47: client may skip advertising task

### LINKED ISSUE #47: Don't require the Client to advertise the task config in every upload request
Currently the Client always advertised the task config in a header. We can save on bandwidth by having the Leader ask the Client for the task config only if it doesn't recognize the task.

We could consider doing something similar for the aggregation and collection flows, but (1) the bandwidth savings are less compelling and (2) we would probably need a signal that the sender (e.g., the Leader in the aggregation flow) is using taskprov.

Originally suggested by @martinthomson on the mailing list.

junyechen1996:
Closes #47.

### DIFF
--- a/draft-wang-ppm-dap-taskprov.md
+++ b/draft-wang-ppm-dap-taskprov.md
@@ -451,29 +451,38 @@ out, it MUST not attempt to upload reports for the task.
 > OPEN ISSUE: In case of opt-out, would it be useful to specify how to report
 > this to the Author?
 
-Once the client opts into a task, it may begin uploading reports for the task.
-Each upload request for that task MUST advertise the task configuration. The
-extension codepoint `taskbind` MUST be offered in the `extensions` field of
-both Leader and Helper's `PlaintextInputShare`. In addition, each report's task
-ID MUST be computed as described in {{definition}}.
+Once the client opts into a task, it may begin uploading reports for the task
+to the Leader. The extension codepoint `taskbind` MUST be offered in the
+`extensions` field of both Leader and Helper's `PlaintextInputShare`. In
+addition, each report's task ID MUST be computed as described in {{definition}}.
+
+The Client SHOULD advertise the task configuration by specifying the encoded
+`TaskConfig` described in {{definition}} in the "dap-taskprov" HTTP header, but
+MAY choose to omit this header in order to save network bandwidth. However, the
+Leader may respond with "unrecognizedTask" if it has not been configured with
+this task. In this case, the Client MUST retry the upload request with the
+"dap-taskprov" HTTP header.
 
 ## Leader Behavior
 
 ### Upload Protocol
 
-Upon receiving a task advertisement from a Client, if the Leader does not
-support this mechanism, it will ignore the HTTP header. In particular, if the
+Upon receiving a Client report, if the Leader does not support the {{taskprov}}
+mechanism, it will ignore the "dap-taskprov" HTTP header. In particular, if the
 task ID is not recognized, then it MUST abort the upload request with
 "unrecognizedTask".
 
-Otherwise, if the Leader does support this mechanism, it first attempts to
-parse the "dap-taskprov" HTTP header payload. If parsing fails, it MUST abort
-with "invalidMessage".
+Otherwise, if the Leader does support this mechanism, it first checks if the
+"dap-taskprov" HTTP header is specified. If not, that means the Client has
+skipped task advertisement. If the Leader recognizes the task ID, it will
+include the client report in the aggregation of that task ID. Otherwise, it
+MUST abort with "unrecognizedTask". The Client will then retry with the task
+advertisement.
 
-Next, it checks that the task ID indicated by the upload request matches the
-task ID derived from the extension payload as specified in
-{{definition}}. If the task ID does not match, then the Leader MUST abort
-with "unrecognizedTask".
+If the Client advertises the task, the Leader checks that the task ID indicated
+by the upload request matches the task ID derived from the extension payload as
+specified in {{definition}}. If the task ID does not match, then the Leader MUST
+abort with "unrecognizedTask".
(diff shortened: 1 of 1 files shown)


draft-wang-ppm-dap-taskprov.md
```diff
@@ -451,11 +451,18 @@ out, it MUST not attempt to upload reports for the task.
...
-ID MUST be computed as described in {{definition}}.
+Once the client opts into a task, it may begin uploading reports for the task
+to the Leader. The extension codepoint `taskbind` MUST be offered in the
+`extensions` field of both Leader and Helper's `PlaintextInputShare`. In
+addition, each report's task ID MUST be computed as described in {{definition}}.
+
+The Client MAY advertise the task configuration by specifying the encoded
+`TaskConfig` {{definition}} in the "dap-taskprov" HTTP header, but may choose
+to just specify the `taskbind` codepoint in the header, in order to save
+network bandwidth. However, the Leader may respond with "unrecognizedTask" if
```
cjpatton, line 462: Hmmm, this is interesting. Why does client need to send the header at all? What do you expect the leader if it has opted in to the task, but the client doesn't send the header?

junyechen1996: The goal is to distinguish between taskprov and non-taskprov tasks. If we don't do this, I guess we can decrypt leader share and try to find the extension codepoint. Both should work. Do you prefer the latter?

cjpatton: If the leader has opted in, then it knows that the task is taskprov. I don't think signaling in the header as well as necessary. Dropping it here would make the protocol simpler, so I'd be in favor of dropping it.

junyechen1996: Sounds good. Updated the PR.

draft-wang-ppm-dap-taskprov.md
```diff
@@ -451,29 +451,38 @@ out, it MUST not attempt to upload reports for the task.
...
-both Leader and Helper's `PlaintextInputShare`. In addition, each report's task
-ID MUST be computed as described in {{definition}}.
+Once the client opts into a task, it may begin uploading reports for the task
+to the Leader. The extension codepoint `taskbind` MUST be offered in the
+`extensions` field of both Leader and Helper's `PlaintextInputShare`. In
+addition, each report's task ID MUST be computed as described in {{definition}}.
+
+The Client MAY advertise the task configuration by specifying the encoded
+`TaskConfig` described in {{definition}} in the "dap-taskprov" HTTP header, but
+may choose to omit this header in order to save network bandwidth. However, the
```
cjpatton, line 461: More perscriptive
```suggestion
The Client SHOULD advertise the task configuration by specifying the encoded
`TaskConfig` described in {{definition}} in the "dap-taskprov" HTTP header, but
MAY choose to omit this header in order to save network bandwidth. However, the
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -451,29 +451,38 @@ out, it MUST not attempt to upload reports for the task.
...
+Leader may respond with "unrecognizedTask" if it has not been configured with
+this task. In this case, the Client MUST retry the upload request by specifying
+the "dap-taskprov" HTTP header.
```
cjpatton, line 464:
```suggestion
this task. In this case, the Client MUST retry the upload request with
the "dap-taskprov" HTTP header.
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -451,29 +451,38 @@ out, it MUST not attempt to upload reports for the task.
...
 ## Leader Behavior
 
 ### Upload Protocol
 
-Upon receiving a task advertisement from a Client, if the Leader does not
-support this mechanism, it will ignore the HTTP header. In particular, if the
-task ID is not recognized, then it MUST abort the upload request with
+Upon receiving a Client report, if the Leader does not support the {{taskprov}}
+mechanism, it will ignore the "dap-taskprov" HTTP header. In particular, if the
+task ID is notrecognized, then it MUST abort the upload request with
```
cjpatton, line 472:
```suggestion
task ID is not recognized, then it MUST abort the upload request with
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -451,29 +451,38 @@ out, it MUST not attempt to upload reports for the task.
...
-Next, it checks that the task ID indicated by the upload request matches the
-task ID derived from the extension payload as specified in
-{{definition}}. If the task ID does not match, then the Leader MUST abort
-with "unrecognizedTask".
+Otherwise, if the Leader does support this mechanism, it first checks if the
+"dap-taskprov" HTTP header is specified. If not, that means the Client has
+skipped task advertisement. If the Leader recognizes the task ID, it will
+include the client report in the aggregation of that task ID. Otherwise, it
+MUST abort with "unrecognizedTask" and ask the Client to retry with the
+"dap-taskprov" HTTP header.
```
cjpatton, line 480: It does not have to "ask" explicitly.
```suggestion
MUST abort with "unrecognizedTask". The Client will then retry with the task advertisement.
```

draft-wang-ppm-dap-taskprov.md
```diff
@@ -451,29 +451,38 @@ out, it MUST not attempt to upload reports for the task.
...
+
+If the Client specifies the "dap-taskprov" HTTP header, the Leader checks that
```
cjpatton, line 482: A simpler, and more complete way to say this is
```suggestion
If the Client advertises the task, the Leader checks that
```

cjpatton (commented):
Looks good, just some minor grammatical stuff

junyechen1996:
Squashed.

cjpatton (approved):

PR #68, opened 2024-08-09
Editorial changes

divergentdave:
This makes a few editorial fixes throughout the document.

- There's one reference to the payload of the extension left over from past drafts (the payload should now be empty).
- There are contradicting statements that the collector receives task parameters via HTTP headers and directly from the task author. I removed mention of HTTP headers, since the collector only initiates HTTP requests.
- I changed places where mention was made of the collector making an upload request or the helper receiving an upload request.


cjpatton (approved):
Thanks, David!

PR #73, opened 2024-10-03
Add salt to task ID computation (*)

### LINKED ISSUE #64: Consider salting the task ID computation
The task ID is just the SHA-256 hash of the `TaskConfig`. Prepending the input with a context string would make cross-protocol attacks on Taskprov less likely.

cjpatton:
Closes #64.

Current the task ID is computed by hashing the raw `TaskConfig`. Prepend the `TaskConfig` with a salt string so that it's less likely we'll confuse the task ID with the hash of the task config in a different context.


junyechen1996 (approved):

cjpatton:
Whoops, I forgot to change base before merging. I cherry-picked this commit onto main and pushed.

Issue #88, opened 2024-10-24
Consider removing the extension point for DP mechanisms


cjpatton:
2024/10/26 interim: It's a little confusing that this draft defines a a registry for DP mechanism codepoints. We could instead move DP mechanisms into a generic task config extension: see https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov/issues/85#issuecomment-2436405240.

cjpatton:
MT's point, which I agree with: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov/issues/85#issuecomment-2436771339

cjpatton:
I've gotten some more feedback offline about this and we're good to remove this. PR incoming.

PR #89, opened 2024-10-25
Add an extension field to `TaskConfig` (*)

### LINKED ISSUE #85: Extensibility
I'm in the process of writing up a few DAP (upload) extensions and realized that if they are going to be generically useful, then it would make sense to have corresponding changes to be included in task configurations.  For this draft, that means having some means to include that configuration in the creation of the task, but there are no extensibility points in this document that would allow that to happen.  Can an extension mechanism be defined?

Most likely, this would have to be like upload extensions, where ignorance of the extension would invalidate the entire request.

I would also recommend that the DP configuration be moved to an extension.  Then you can move the definition of the specifics to the DP draft and remove one potentially-awkward interdependency.

cjpatton:
Closes #85. We'll consider #88 (whether to remove the DP mechanism and suggest implementations do this in an extension) in a future PR.

Notable features:

* Create a registry for new taskbind extensions
* Make extensions mandatory-to-implement when provisioning a task

cc @martinthomson

### DIFF
--- a/draft-ietf-ppm-dap-taskprov.md
+++ b/draft-ietf-ppm-dap-taskprov.md
@@ -134,7 +134,7 @@ the report extensions for each Aggregator as described in {{Section 4.4.3 of
 ~~~
 enum {
     taskbind(0xff00),
-    (65535)
+    (2^16-1)
 } ExtensionType;
 ~~~
 
@@ -197,8 +197,8 @@ struct {
     /* The batch mode and its parameters. */
     opaque batch_config<1..2^16-1>;
 
-    /* Time from which Clients will start uploading reports to this
-    task. */
+    /* The earliest timestamp that will be accepted for this task. */
+       task. */
     Time task_start;
 
     /* The duration of the task. */
@@ -209,16 +209,26 @@ struct {
 
     /* Determines the VDAF type and its config parameters. */
     opaque vdaf_config<1..2^16-1>;
+
+    /* Taskbind Extensions. */
+    TaskbindExtension extensions<0..2^16-1>;
 } TaskConfig;
 ~~~
 
 The purpose of `TaskConfig` is to define all parameters that are necessary for
-configuring each party. It includes all the fields to be associated with a
-task. It also includes an opaque `task_info` field that is specific to a
-deployment. For example, this can be a string describing the purpose of this
-task. It does not include cryptographic assets shared by only a subset of the
-parties, including the secret VDAF verification key {{!VDAF}} or public HPKE
-configurations {{!RFC9180}}.
+configuring each party. It includes all parameters listed in {{Section 4.3 of
+!DAP}} as well as two additional fields:
+
+* `task_info` is an opaque field whose contents are specific to the deployment.
+  For example, this might be a human-readable string describing the purpose of
+  this task.
+
+* `extensions` is a list of extensions to this document. The format and
+  semantics of extensions are describe in {{taskbind-extensions}}.
+
+This structure does not include cryptographic assets shared by only a subset of
+the parties, including the secret VDAF verification key {{!VDAF}} or public
+HPKE configurations {{!RFC9180}}.
 
 The `batch_config` field defines the DAP batch mode. Its contents are as follows:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap-taskprov.md
```diff
@@ -198,7 +198,7 @@ struct {
     opaque batch_config<1..2^16-1>;
 
     /* Time from which Clients will start uploading reports to this
```
martinthomson, line 200:
```suggestion
    /* The earliest timestamp that will be accepted for this task. */
```

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -209,16 +209,26 @@ struct {
...
-deployment. For example, this can be a string describing the purpose of this
-task. It does not include cryptographic assets shared by only a subset of the
-parties, including the secret VDAF verification key {{!VDAF}} or public HPKE
-configurations {{!RFC9180}}.
+configuring each party. It includes all parameters listed in {{Section 4.3 of
+!DAP}} as well as two additional fields:
+
+* `task_info` is an opaque field whose contents are specific to the deployment.
+  For example, this might be a human-readable string describing the purpose of
+  this task.
```
martinthomson, line 224: This smells like an extension to me.

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -301,6 +311,34 @@ unimplemented VDAF).
...
+  (2^16-1)
+} ExtensionType;
+~~~
+
+The `extension_type` identifies the extension and `extension_data` is
+structured as specified by the extension.
+
+Extensions are treated as mandatory-to-implement in the protocol described in
+{{taskprov}}. In particular protocols participants MUST opt-out of tasks
+containing unrecognized extensions. See {{provisioning-a-task}}.
```
martinthomson, line 340: The question I have is how codepoints here relate to codepoints in DAP, if that makes sense.

If I define an extension 0xAAAA here does that match the 0xAAAA codepoint in DAP?

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -626,6 +668,28 @@ Name:
...
+This registry contains the following columns:
+
+Value:
+: The two-byte identifier for the extension
+
+Name:
+: The name of the extension
+
+Reference:
+: Where the eechanism is defined
```
wangshan, line 684:
```suggestion
: Where the mechanism is defined
```

draft-ietf-ppm-dap-taskprov.md
wangshan: +1

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -301,6 +311,34 @@ unimplemented VDAF).
...
+
+~~~ tls-presentation
+struct {
+  ExtensionType extension_type;
+  opaque extension_data<0..2^16-1>;
+} Extension;
+
+enum {
+  reserved(0),
+  (2^16-1)
```
wangshan, line 331: nit: the doc mentions DAP's ExtensionType which uses 65535, we should make it consistent, probably use 2^16-1 everywhere since that's what the other enum definition uses in this doc.

draft-ietf-ppm-dap-taskprov.md
cjpatton: I'm not sure it should be if it's mandatory to implement. In any case, I think we should discuss removing the task_info field in a separate issue.

draft-ietf-ppm-dap-taskprov.md
wangshan: I think this makes sense, do we want to add some text to explain this strategy. I can see this same question coming up in the future.

wangshan (approved):
LGTM.
@cjpatton consider explaining the extension code point strategy in the text (perhaps in the main DAP text and refer to it from here)

draft-ietf-ppm-dap-taskprov.md
cjpatton: Yup, will do.

cjpatton: Done.

cjpatton:
Squashed (so that I can start stacking on this PR).

PR #91, opened 2024-10-30
Improve extension points for batch modes and VDAFs (*)

cjpatton:
Stacked on #90.

Along the same lines as
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/627

Harmonize the extension points for new batch modes or VDAFs with
Tasbind extensions by following the Tag-Length-Value (TLV) pattern. That
is, encode the variant followed by length-prefixed payload, the format
of which is determined by the variant. This is inteded to make it easier
to extend the document in the future.

This change also has a couple of bug fixes:

* Fix encoding of Prio3Sum config (the circuit now requires a maximum
  measurement rather than the bit length of the maximum measurement).
* Add encoding for Prio3MultihotCountVec
﻿﻿

### DIFF
--- a/draft-ietf-ppm-dap-taskprov.md
+++ b/draft-ietf-ppm-dap-taskprov.md
@@ -178,6 +178,8 @@ ID of the request.
 The task configuration is encoded as follows:
 
 ~~~
+uint32 VdafType; /* As defined in Section 10 of [VDAF] */
+
 struct {
     /* Info specific for a task. */
     opaque task_info<1..2^8-1>;
@@ -195,6 +197,7 @@ struct {
     uint32 min_batch_size;
 
     /* The batch mode and its parameters. */
+    BatchMode batch_mode;
     opaque batch_config<1..2^16-1>;
 
     /* The earliest timestamp that will be accepted for this task. */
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap-taskprov.md
```diff
@@ -686,8 +716,8 @@ that define:
 1. A new Taskbind extension
 
 Documents defining either a new DAP batch mode or VDAF SHOULD include a section
-titled "Taskbind Considerations" that specifies how to extend the `BatchConfig`
-structure and the `VdafConfig` structure respectively.
+titled "Taskbind Considerations" that specifies the payload of
+`TaskConfig.batch_config` or `TaskConfig.vdaf_config` respectively.
```
cjpatton, line 720: This change is the main pay off of this PR. "Extend the `VdafConfig` structure" is not very well defined.

wangshan (approved):

draft-ietf-ppm-dap-taskprov.md
junyechen1996: So do we have any language on what if `batch_mode` or `vdaf_type` is unrecognized by an aggregator? (which is the initial motivation to make `vdaf_config` an opaque)

junyechen1996 (approved):

draft-ietf-ppm-dap-taskprov.md
cjpatton: Yes, if not recognized then the party MUST opt out.

PR #95, opened 2024-11-22
Tweak lower bounds on field lengths to 0 instead of 1.

branlwyd:
Both of the relevant fields can include zero-length values in some cases, so this change is required for taskprov to be implementable (while still following the specification).

Also, clean up a comment typo.


cjpatton (approved):
👍  thanks!

PR #96, opened 2024-11-25
Fix references

divergentdave:
This fixes a few editorial issues in references.

* One of the IANA considerations sections had a reference to VDAF that should instead be a reference to DAP.
* There was a reference to `{{Section C of !VDAF}}`. Appendix C is the test vectors. Based on context, I think this should have been a reference to the Prio3 section, so I changed it accordingly.
* I updated invalid section numbers pointing to the section of DAP discussing extensions.


cjpatton (approved):
❤️ !

Issue #94, opened 2024-11-21
Provide guidance on whether to use public vs private extension for Taskbind


branlwyd:
DAP-13 introduced the concept of "[public extensions](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/632)", complementing the preexisting (private) extension mechanism.

These extensions can conflict with one another (i.e. it is not possible to send the same extension type in both the public & private extensions), but it is left open if/how the choice of public vs private extensions changes the semantics of any given extension.

I think taskprov should provide some guidance as to which type of extension the Taskbind extension should be used with.

My initial thoughts are: the Taskbind extension uses an empty extension body, and is purposefully included in the extensions for both the Leader & Helper aggregators. Therefore, I think recommending this extension to be included in the public extensions makes sense. I don't have a strong opinion on whether we should block including this extension in the private extensions.

branlwyd:
Ah, actually, the [Client Behavior](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-taskprov-01.html#name-client-behavior) section says the following:

> The extension codepoint taskbind MUST be offered in the extensions field of both Leader and Helper's PlaintextInputShare. In addition, each report's task ID MUST be computed as described in [Section 3](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-taskprov-01.html#definition).

This implies that the extension must be included in the _private_ extensions, since private extensions are included in each aggregators' `PlaintextInputShare`.

I suggest we at least accept public extensions, too, and probably recommend the use of a public extension.

PR #98, opened 2024-12-11
Allow taskbind extension to be included in public report extensions.

branlwyd:
The extension can also be included in the private extensions; standard DAP semantics will disallow it from being included in both extension mechanisms.


branlwyd:
Closes #94.

cjpatton (approved):

PR #105, opened 2025-08-28
Add text to disallow TaskConfig containig repeated TaskprovExtension …

wangshan:
…types

Resolves issue https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov/issues/100

### DIFF
--- a/draft-ietf-ppm-dap-taskprov.md
+++ b/draft-ietf-ppm-dap-taskprov.md
@@ -383,6 +383,10 @@ Extensions are treated as mandatory-to-implement in the protocol described in
 {{taskprov}}. In particular, protocols participants MUST opt-out of tasks
 containing unrecognized extensions. See {{provisioning-a-task}}.
 
+If the same extension type appears more than once among the Taskprov extensions
+in the `TaskConfig` structure, then protocols participants MUST opt-out of tasks
+with repeated extensions.
+
 Note that Taskprov extensions are semantically distinct from DAP report
 extensions and do not share the same codepoint registry
 ({{taskprov-extension-registry}}). Future documents may want to define both a
@@ -509,6 +513,8 @@ A protocol participant MUST opt out if:
 1. The DAP batch mode or VDAF is not implemented.
 
 1. One of the extensions is not recognized.
+
+1. One of the extension types appear more than once.
 
 The behavior of each protocol participant is determined by whether or not they
 opt in to a task.


draft-ietf-ppm-dap-taskprov.md
```diff
@@ -611,7 +617,7 @@ Finally, the Helper completes the request as usual, deriving the VDAF
 verification key for the task as described in {{vdaf-verify-key}}. During an
 aggregation job, for any report share that does not include the `taskprov`
 extension with an empty payload, the Helper MUST mark the report as invalid
-with error "invalid_message" and reject it.
+with error "invalidMessage" and reject it.
```
cjpatton, line 620: This is actually correct as is. Here we mean that the `ReportError` should be set to the `invalid_message` variant. `invalidMessage` refers to an error variant that would be stuffed in a JSON response when aborting a DAP request.

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -383,6 +383,10 @@ Extensions are treated as mandatory-to-implement in the protocol described in
 {{taskprov}}. In particular, protocols participants MUST opt-out of tasks
 containing unrecognized extensions. See {{provisioning-a-task}}.
 
+If the same extension type appears more than once among the Taskprov extensions
+in the `TaskConfig` structure, then protocols participants MUST opt-out of tasks
+with repeated extensions.
```
cjpatton, line 388: To have parity with report extension handling, we should treat the entire `TaskConfig` structure as invalid. That would mean aborting DAP rather than just opting out. Likewise below.

wangshan: I think opt-out will have the same effect, because for both Leader and Helper, if they decide to opt-out, the result is abort DAP with invalidTask:
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov/blob/d96bb5a98e424a03d40e2a7608e518c659cbef14/draft-ietf-ppm-dap-taskprov.md?plain=1#L558
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov/blob/d96bb5a98e424a03d40e2a7608e518c659cbef14/draft-ietf-ppm-dap-taskprov.md?plain=1#L607

cjpatton: Fair point! But I think this should be handled as `invalidMessage` instead, since we're saying that it's some how "gramatically incorrect" to include the same code point multiple times.

wangshan: It only makes sense for invalidMessage if it's provisioned in-band. But this section describes the overall behavior of a bad taskConfig, it might happen out-of-band, when no Reports have been seen yet.

cjpatton: Yes, good point!

cjpatton (approved):

PR #106, opened 2025-08-29
Add considerations for task enumeration attacks

### LINKED ISSUE #102: Task enumeration attacks
`taskbind` defines a means of deterministically deriving a DAP task ID from an [encoding of the task](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-taskprov-01#section-3.1). That encoding includes things like the VDAF configuration as well as identifiers for the participating aggregators. I wonder if this makes it possible to enumerate taskprov tasks supported by an aggregator, as well as partner aggregators.

If I can guess at a pair of aggregator URLs and a VDAF config, then I can construct a report and upload it to the task leader and see if the upload gets accepted, which reveals that I guessed correctly. There are quite a few fields on `TaskConfig`, but the space to explore isn't THAT big, especially since I might have insight into what VDAF is in use (perhaps because a software vendor might publish some details about how they're using DAP) and what the likely set of aggregators that might be participating is.

Using request authentication is a partial mitigation here: presumably if an HTTP client can authenticate to the aggregator, there's a measure of trust implied there. But imagine a setting where one aggregator uses taskprov with two different partners. You want to allow each partner to dynamically configure tasks, but you don't want to let them walk the task space and learn about your other partner.

I think this merits at least a mention in security considerations (alongside the existing discussion of the DoS attack where storage gets flooded with spurious task definitions), but a more robust fix would be to add a field to `TaskConfig` that is REQUIRED to be, let's say, 32 bytes of randomness. That would make it impossible to guess valid `TaskConfig`s.

cjpatton:
Closes #102.

cc @tgeoghegan

### DIFF
--- a/draft-ietf-ppm-dap-taskprov.md
+++ b/draft-ietf-ppm-dap-taskprov.md
@@ -668,6 +668,21 @@ SHOULD arrange for the Author to digitally sign the task configuration so that
 Clients cannot forge task creation, e.g., via an extension to Taskprov
 ({{taskprov-extensions}}).
 
+Support for the Taskprov extension may render a deployment of DAP more
+susceptible to task enumeration attacks ({{Section 8.6.1 of !DAP}}). For
+example, if the Leader's upload endpoint is unauthenticated, then any HTTP
+client can learn if a Leader supports a particular task configuration by
+uploading a report for it with the Taskprov extension. Aggregators can mitigate
+these kinds of attack by:
+
+1. Requiring authentication of all APIs, including the upload endpoint (see
+   {{Section 3.3 of !DAP}});
+
+1. Enforcing rate limits on unauthenticated APIs; or
+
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap-taskprov.md
```diff
@@ -668,6 +668,21 @@ SHOULD arrange for the Author to digitally sign the task configuration so that
 Clients cannot forge task creation, e.g., via an extension to Taskprov
 ({{taskprov-extensions}}).
 
+Support for the Taskprov extension may render a deployment of DAP more
+susceptible to task enumeration attacks ({{Section 8.6.1 of !DAP}}). For
+example, if the Leader's upload enpdoint is unauthenticated, then any HTTP
```
wangshan, line 673:
```suggestion
example, if the Leader's upload endpoint is unauthenticated, then any HTTP
```

wangshan (approved):

PR #122, opened 2026-01-05
Move task times to an extension

martinthomson:
As discussed, constraining the time of a task is something that was intended to drive key rotation, but other mechanisms are better suited to that and having a requirement to configure all clients with time information makes it more operationally challenging to use this.

Having the extension makes the decision to bind to a particular time range optional.

Note that this doesn't get rid of time precision; that's something that needs to be agreed (or fixed).  In updating this, I'm banking on @tgeoghegan's change in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/759 landing with a definition of `TimePrecision` that matches what I've removed here.

### DIFF
--- a/draft-ietf-ppm-dap-taskprov.md
+++ b/draft-ietf-ppm-dap-taskprov.md
@@ -261,12 +261,6 @@ struct {
     BatchMode batch_mode;
     opaque batch_config<0..2^16-1>;
 
-    /* The earliest timestamp that will be accepted for this task. */
-    Time task_start;
-
-    /* The duration of the task. */
-    Duration task_duration;
-
     /* Determines the VDAF type and its config parameters. */
     VdafType vdaf_type;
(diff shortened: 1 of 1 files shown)


wangshan:
we can discuss in DAP too, but just want to point out that the decision to have task start and duration is not just for key rotation, because key rotation can be independent of tasks. The original goal is to allow a time bounded task, some if not most collections should have a clear start and end. I believe this is also the spirit of DAP. @cjpatton @martinthomson

wangshan (commented):
As discussed in the Jan 20, 2026 meeting we agreed to 1). close this PR, 2). figure out how to move taskprov to core DAP, 3). Martin to recreate the PR once it's moved.

martinthomson:
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/776 covers this now.

Issue #125, opened 2026-01-30
Url is not defined


martinthomson:
The draft [presently says](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-taskprov#section-3.1-13):

> The definition of Time, Duration, Url, and BatchMode follow those in [[DAP](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap-taskprov/draft-ietf-ppm-dap-taskprov.html#DAP)].

However, I can't find a definition of `Url` in DAP.

I'm guessing that this is something simple like this:

```tls-syntax
opaque Url<1..2^16-1>;
```

It would be good to have that defined.  (This probably applies to DAP now that the encoding of task configuration is moving, but I'm filing it here for better visibility.)

cjpatton:
Yup, that's how it's intended to be defined. I believe at one point it was, in DAP.

cjpatton:
No longer relevant.

Issue #120, opened 2025-12-16
Use `Interval` from draft-ietf-ppm-dap to represent task lifetime


cjpatton:
Time intervals are are represented as a start time (UNIX timestamp) plus the number of `time_precision`. That is, the end time is `start_time + duration*time_precision`. We can use `Interval` to represent this data in `TaskConfig`. We should also clarify how `duration` is interpreted.

tgeoghegan:
[DAP 4.1.1](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-16#section-4.1.1) defines an interval as:

```
struct {
  Time start;
  Duration duration;
} Interval;
```

And both `Time` and `Duration` are defined as "integers representing a number of time_precisions" so I think it follows from those definitions. But that doesn't mean we can't make it even more clear. Maybe we should put some examples of `Time`, `Duration` and `Interval` values in DAP next to the type definitions.

cjpatton:
No longer relevant since we don't define the task config structure anymore.

Issue #115, opened 2025-12-10
Extension model for VDAFs and batch modes


martinthomson:
DAP defines identifiers for both VDAFs and batch modes.  When new values are defined for those, Taskprov cannot be used until a definition is given for the values that are encoded into `vdaf_config` and `batch_config`, respectively.

This is basically OK, but it imposes a constraint on extensions.  This is current spelled out with:

> Documents defining either a new DAP batch mode or VDAF SHOULD include a section titled "Taskprov Considerations" that specifies the payload of TaskConfig.batch_config or TaskConfig.vdaf_config respectively.

This is a recommendation, but I would make this draft update the registries for these and request that new definitions identify the specification of the encoding for parameters.  Otherwise, you might end up in bad situations.

Consider that someone chooses to ignore the "SHOULD" here and doesn't define how to encode the configuration for a new extension (of either type).  That might be due to an expectation that Taskprov is not needed.  Later, someone decides to define that.  They do that by writing down how to do that in a document of their own, probably an internet-draft.

But what if two someones define an encoding and they concurrently reach different conclusions about the encoding?  Interoperability failure.

Adding a pointer to the respective registries means that you force a definition to consider Taskprov.  And in doing so, you ensure that encodings are considered for all extensions to these.  The cost is that someone who decides not to define an encoding will prevent the use of Taskprov.  At least until their specification -- and the IANA registry entry -- are updated, but it becomes clear that only one such encoding can exist.

So the "SHOULD" becomes "MUST either define a canonical encoding for related configuration or note that Taskprov cannot be used with the {VDAF|batch mode}".  "SHOULD" is almost always the wrong choice anyway.

Issue #114, opened 2025-12-10
What purpose does the Taskprov report extension serve?


martinthomson:
This extension is in the title of [Section 3](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-taskprov#section-3), but I can't work out what value is served by having it present.

I can see why you might think that an extension is necessary so that both client and aggregators agree that Taskprov is in effect, but I don't think that's right.

There are three outcomes possible when a report is submitted without this extension:
1. The server doesn't support Taskprov, in which case the task ID will be unknown to it. There is no way to recover from this state and that's OK.
2. The server supports Taskprov, but doesn't know about the task ID (yet).  This is where the provisioning process in Section 4 might come into play.
3. The server already knows about the task (with or without Taskprov, this is mostly fine).

Because the aggregators need to use Taskprov to generate a task configuration (and task ID), there is no situation where they are not aware of the task being either created using Taskprov or not.

As far as I can see, the most important potential reason to define a report extension doesn't hold in this case.  That is, it doesn't help ensure that clients and servers agree on something.  Clients do not need to care how the task was configured.  Either they are told that Taskprov was involved or not.  In either case, they need to know about the task configuration and how it relates to the task ID, but there is no substantive difference for clients operating in two modes, assuming that the task ID is the same:

a. A client that is given the task configuration and task ID, as per the original DAP.
b. A client that is given the task configuration and told to generate a task ID using the process here.

The client needs to know and trust the aggregators.  Taskprov creates a stronger binding between a given task configuration and a concrete task, but if the original DAP can be secure on the basis that the client trusts that the configuration the aggregators are using matches its own, then it is also true when using Taskprov.  Taskprov provides stronger assurances about that correspondence, but this is only relevant if the original DAP is vulnerable to attacks where clients are given the wrong configuration for a task.  I don't believe that to be the case, but if it were true, then the entire task ID design in DAP is broken.  More on that in a separate thread.

The only reason I can imagine being an issue related to the report extension is much more about the details of how the VDAF is executed when Taskprov is in play. That is because Taskprov fixes the verification key derivation. I'll open a follow-up issue on that part.

cjpatton:
I'm not sure I follow completely. You write:

[...]

PR #126, opened 2026-03-02
Remove task parameter binding from the draft

tgeoghegan:
PPM WG consensus is that we should do task configuration binding/commitment in DAP itself. This is realized in [1]. This PR on taskprov mostly deletes content that now lives in draft-ietf-ppm-dap, leaving the in-band task provisioning mechanism. A couple things to call out:

- numerous portions of this document need to refer to DAP's discussion of task configuration, but since no draft including that is yet published, I put in a dummy section titled "TODO fix dap reference" and refer to that elsewhere. Once draft-ietf-ppm-dap-18 is released, we can replace those with specific section references, but for now, this gets us a buildable document we can review. A subsequent PR will replace these with up-to-date references to specific sections of draft-ietf-ppm-dap-18.
- Since the task configuration mechanism in DAP also has extensions, taskprov doesn't need to define its own extension mechanism. And to my reading, the entire section on extending taskprov can be deleted, as its considerations now apply to DAP itself.
- The taskprov *report* extension is replaced with a *task* extension. This achieves the same goal of ensuring all protocol participants agree that taskprov is in use, but avoids making every report bigger.

[1]: ietf-wg-ppm/draft-ietf-ppm-dap#774

### DIFF
--- a/draft-ietf-ppm-dap-taskprov.md
+++ b/draft-ietf-ppm-dap-taskprov.md
@@ -59,21 +59,18 @@ parties that this mechanism was used to configure the task.
 (RFC EDITOR: Remove this paragraph.) This draft is maintained in
 https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov.
 
-> TODO: update this DAP reference once draft-ietf-ppm-dap-18 is out
+The Distributed Aggregation Protocol {{!DAP=I-D.draft-ietf-ppm-dap-18}} enables
+secure aggregation of a set of reports submitted by Clients. This process is
+centered around a "task" that determines, among other things, the cryptographic
+scheme to use for the secure computation (a Verifiable Distributed Aggregation
+Function {{!VDAF=I-D.draft-irtf-cfrg-vdaf-15}}), how reports are partitioned
+into batches, and privacy parameters such as the minimum size of each batch.
 
-The DAP protocol {{!DAP=I-D.draft-ietf-ppm-dap-16}} enables secure aggregation
-of a set of reports submitted by Clients. This process is centered around a
-"task" that determines, among other things, the cryptographic scheme to use for
-the secure computation (a Verifiable Distributed Aggregation Function
-{{!VDAF=I-D.draft-irtf-cfrg-vdaf-15}}), how reports are partitioned into
-batches, and privacy parameters such as the minimum size of each batch. See
-{{todo-fix-dap-reference}} for a complete listing.
-
-{{taskprov}} specifies one possible mechanism for provisioning DAP tasks that
-is built on top of the task configuration definition in
-{{todo-fix-dap-reference}}. Its chief design goal is to make task configuration
-completely in-band, via HTTP request headers. It also defines a task extension
-that signals to parties that this mechanism was used to configure the task.
+{{taskprov}} specifies one possible mechanism for provisioning DAP tasks that is
+built on top of the task configuration definition in {{Section 4.2 of !DAP}}.
+Its chief design goal is to make task configuration completely in-band, via HTTP
+request headers. It also defines a task extension that signals to parties that
+this mechanism was used to configure the task.
 
 ## Change Log
 
@@ -166,29 +163,29 @@ Task author:
 
 When Taskprov is in use, protocol participants include a `Taskprov` extension
 (see {{task-extension}}) in the task's `TaskConfiguration` structure as
-described in {{todo-fix-dap-reference}}.
+described in {{Section 4.2 of !DAP}}.
 
 The payload of the extension MUST be empty. If the payload is non-empty, then
 the protocol participant MUST refuse to participate in the task.
 
-When the client uses the Taskprov extension, it computes the task ID
-({{todo-fix-dap-reference}}) as follows:
+When the client uses the Taskprov extension, it computes the task ID ({{Section
+4.2 of !DAP}}) as follows:
 
 ~~~
 task_id = SHA-256(SHA-256("dap-taskprov task id") || task_config)
 ~~~
 
-where `task_config` is a `TaskConfiguration` structure defined in
-{{todo-fix-dap-reference}}. Function SHA-256() is as defined in {{SHS}}.
+where `task_config` is a `TaskConfiguration` structure. Function SHA-256() is as
+defined in {{SHS}}.
 
 The task configuration is bound to each report share (via HPKE authenticated and
-associated data, see {{todo-fix-dap-reference}}). Deterministically deriving the
+associated data, see {{Section 4.4.2.1 of !DAP}}). Deterministically deriving the
 task ID from the encoded `TaskConfiguration` ensures that Clients, Aggregators
 and the Collector can only agree on the task ID if they also agree on the
 parameters, including whether Taskprov is in use. This is accomplished by the
 following Aggregator behavior.
 
-During aggregation ({{Section 4.6 of !DAP}}) in a task where Taskprov is in use,
+During aggregation ({{Section 4.5 of !DAP}}) in a task where Taskprov is in use,
 each Aggregator processes reports as follows.
 
 First, it looks up the ID and parameters associated with the task. Note the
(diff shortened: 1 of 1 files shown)


tgeoghegan:
@cjpatton Don't know why but you didn't get tagged as a reviewer here.

draft-ietf-ppm-dap-taskprov.md
cjpatton: This is no longer true: agreement on the task ID is no longer necessary to agree on the parameters.

cjpatton (commented):
All the report extension does now is ensure the task ID is derived deterministically from the task config. The intent was to bind the task config to report/aggregate shares via the task ID, but that's now accomplished by binding directly to the task config.

We could remove the report extension from the draft and leave just the task provisioning mechanism. We would need to figure out how to pick the ID: we could simply allow the Author to pick it and encode it alongside the config in the task advertisement header. This seems like it would be strictly more useful than what we have now.

draft-ietf-ppm-dap-taskprov.md
martinthomson:
```suggestion
The task ID is generated deterministically from the encoded `TaskConfiguration`,
ensuring that independent Clients are able to generate the same task ID.
Task IDs are generated following the process below.
```

Note that it would be helpful if, rather than "below", there was a section to reference.

martinthomson: As discussed 2026-03-27: This can go.  See #114 for rationale.  And #39 for why this was originally added (since fixed in core DAP).

tgeoghegan: To amend this: we discussed further in #ppm on ietf.slack.com and concluded that it's valuable for everyonein teh protocol to agree on whether taskprov is in use at all. To that end, we moved the report extension to a task extension.

tgeoghegan: I believe I've rewritten the paragraph in a manner that's satisfactory, but please suggest further changes if not.

tgeoghegan: I believe I've rewritten the paragraph in a manner that's satisfactory, but please suggest further changes if not. The aggregator behavior in question is in the paragraph immediately after this one so a section reference feels redundant. I rephrased as "following Aggregator behavior" which I'm not sure is amazing either.

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -518,8 +295,8 @@ with `SHA-256()` as defined in {{SHS}}.
 ## Opting into a Task {#provisioning-a-task}
 
 Prior to participating in a task, each protocol participant must determine if
-the `TaskConfig` disseminated by the Author can be configured. The participant
-is said to "opt in" to the task if the derived task ID (see
+the `TaskConfigration` disseminated by the Author can be configured. The
```
tgeoghegan: Two questions here for the document's editors:

- Is "protocol participant" on L297 correct? The whole point of taskprov is that the aggregators don't interact with the task Author. Should this just be the Client? Or maybe Client and Collector?
- Is it excessively prescriptive to assert that the Author has to disseminate a `TaskConfiguration` structure to Clients? The Author<->Client interactions are out of scope of this document, aren't they?

tgeoghegan:
Updated this PR based on discussion yesterday. The headline change here is that the taskprov report extension is now a task extension. It serves the same purpose, of ensuring that all participants agree that the taskprov extension is in use. But this way each report is smaller.

Besides these changes, one other question for the document's owners: what do you want to do about the dangling `{{todo-fix-dap-reference}}` references? Should we block this PR on draft-ietf-ppm-dap-18 being published, or merge this and circle back and fix those references in a subsequent PR? Note that there's numerous out of date section references in the document anyway.

This PR will close issues:

- ~#114~ 
- #125 (the `Url` type is no longer referenced)
- #123 (taskprov extensions don't exist anymore)
- #120 (`task_interval` extension is now in draft-ietf-ppm-dap)
- #115 (draft-ietf-ppm-dap discusses extensibility requirements)

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -442,7 +220,7 @@ is incumbent on all protocol participants to verify the task configuration
 disseminated by the Author and opt-out if the parameters are deemed insufficient
 for privacy. In particular, adopters of this mechanism should presume the
 Author is under the adversary's control. In fact, we expect in a real-world
-deployment that the Author may be co-located with the Collector.
+deployment that the Author may be co-located with the Collector.
```
martinthomson, line 223: This whitespace change is troubling.  tab to space or space to tab?  Or something else?

martinthomson (commented):
This doesn't establish why the task extension is needed.  That seems pretty important because [I can't see what value it provides](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov/issues/114).

draft-ietf-ppm-dap-taskprov.md
tgeoghegan: It looks like this used to be `c2a0`, a non-breaking space. Now it's just a space.

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -518,8 +296,8 @@ with `SHA-256()` as defined in {{SHS}}.
 ## Opting into a Task {#provisioning-a-task}
 
 Prior to participating in a task, each protocol participant must determine if
-the `TaskConfig` disseminated by the Author can be configured. The participant
-is said to "opt in" to the task if the derived task ID (see
+the `TaskConfigration` disseminated by the Author can be configured. The
```
wangshan:
```suggestion
the `TaskConfiguration` disseminated by the Author can be configured. The
```

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -636,25 +404,23 @@ First, the Helper attempts to parse payload of the "DAP-Taskprov" HTTP header.
...
-whether to opt into the task as described in {{provisioning-a-task}}. If the
-Collector opts out, it MUST NOT attempt to initialize collection jobs for the
-task.
+Upon receiving a `TaskConfiguration` from the Author, the Collector first
+decides whether to opt into the task as described in {{provisioning-a-task}}. If
+the Collector opts out, it MUST NOT attempt to initialize collection jobs for
+the task.
 
 Otherwise, once opted in, the Collector MAY begin to issue collect requests for
 the task. The task ID for each request MUST be derived from the `TaskConfig` as
```
wangshan:
```suggestion
the task. The task ID for each request MUST be derived from the `TaskConfiguration` as
```

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -46,11 +46,7 @@ informative:
 
 --- abstract
 
-An extension for the Distributed Aggregation Protocol (DAP) is specified that
-cryptographically binds the parameters of a task to the task's execution. In
-particular, when a client includes this extension with its report, the servers
-will only aggregate the report if all parties agree on the task parameters.
-This document also specifies an optional mechanism for in-band task
+An extension for the Distributed Aggregation Protocol (DAP) for in-band task
 provisioning that builds on the report extension.
```
cjpatton:
```suggestion
This document defines a mechanism for provisioning tasks for the Distributed Aggregation Protocol (DAP). The task configuration is provisioned in-band via an HTTP header. This document also specifies a task extension that signals to parties that this mechanism was used to configure the task.
```

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -61,54 +57,34 @@ provisioning that builds on the report extension.
...
+{{todo-fix-dap-reference}} for a complete listing.
 
 {{taskprov}} specifies one possible mechanism for provisioning DAP tasks that
-is built on top of the extension in {{definition}}. Its chief design goal is to
-make task configuration completely in-band, via HTTP request headers. Note that
-this mechanism is an optional feature of this specification; it is not required
-to implement the DAP report extension in {{definition}}.
+is built on top of the task configuration definition in
+{{todo-fix-dap-reference}}. Its chief design goal is to make task configuration
+completely in-band, via HTTP request headers.
```
cjpatton, line 74: Also mention the task extension here

cjpatton (approved):
I'd like to clean this up a bit, but I think we should merge this first. I'll do a pass of the doc and put up a PR.

tgeoghegan:
@cjpatton just to be clear, you're OK with taking this change including the `{{todo-fix-dap-reference}}`s?

cjpatton:
> @cjpatton just to be clear, you're OK with taking this change including the `{{todo-fix-dap-reference}}`s?

I can go either way.

tgeoghegan:
OK, I'll let you and the other taskprov editors consider that question. I've rebased this PR so as far as I'm concerned it's ready to merge.

tgeoghegan:
I added a commit to fix up section references for draft-ietf-ppm-dap-18. This should be OK to merge now.

wangshan (approved):

PR #129, opened 2026-07-03
Remove the task extension altogether

### LINKED ISSUE #34: Error type registration
The document currently suggests that the type `urn:ietf:params:ppm:dap:error:invalidTask` should be used in problem details responses. However, DAP says "Servers MUST NOT use the DAP URN namespace for errors not listed in the appropriate IANA registry". Thus, the taskprov document's 'IANA Considerations' section should either add an `invalidTask` entry to the contents of DAP's URN registry, or create a separate namespace of its own.

### LINKED ISSUE #114: What purpose does the Taskprov report extension serve?

cjpatton:
Closes #34.
Closes #114.

Continue to derive the task ID from the config, ensuring that reports are never processed with different configs in the same task.

While at it:

1. Complete IANA considerations for the new DAP error type.

2. Demote HPKE to an informative reference.

### DIFF
--- a/draft-ietf-ppm-dap-taskprov.md
+++ b/draft-ietf-ppm-dap-taskprov.md
@@ -47,10 +47,12 @@ informative:
 --- abstract
 
 This document defines a mechanism for provisioning tasks for the Distributed
-Aggregation Protocol (DAP). The task configuration is provisioned in-band via an
-HTTP header. This document also specifies a task extension that signals to
-parties that this mechanism was used to configure the task.
-
+Aggregation Protocol (DAP). The parameters of the task being executed are
+piggy-backed on each request, enabling deployment scenarios in which the
+process of provisioning these parameters is fully automated. Long-lived
+cryptographic assets are still exchanged out-of-band: accordingly, this
+document also specifies a mechanism for deriving a per-task VDAF verification
+key from a pre-shared secret.
 
 --- middle
 
@@ -66,11 +68,15 @@ scheme to use for the secure computation (a Verifiable Distributed Aggregation
 Function {{!VDAF=I-D.draft-irtf-cfrg-vdaf-15}}), how reports are partitioned
 into batches, and privacy parameters such as the minimum size of each batch.
 
-{{taskprov}} specifies one possible mechanism for provisioning DAP tasks that is
-built on top of the task configuration definition in {{Section 4.2 of !DAP}}.
-Its chief design goal is to make task configuration completely in-band, via HTTP
-request headers. It also defines a task extension that signals to parties that
-this mechanism was used to configure the task.
+This document specifies a mechanism for provisioning DAP tasks that is built on
+top of the task configuration definition in {{Section 4.2 of !DAP}}. Its chief
+design goal is to make task configuration completely in-band, via HTTP request
+headers. Long-level cryptographic assets, such as HPKE configurations
+{{?RFC9180}} and VDAF verification keys {{Section 5.2 of !VDAF}}, are presumed
+to be established out-of-band. Accordingly, this document specifies a mechanism
+for deriving a per-task verification key from a pre-shared secret in a manner
+that satisfies the security requirements for this key ({{Section 9.1 of
+!VDAF}}).
 
 ## Change Log
 
@@ -78,17 +84,12 @@ this mechanism was used to configure the task.
 
 (\*) Indicates a change that breaks wire compatibility with the previous draft.
 
-05:
-
-- Task binding is moved to {{!DAP}}
-  (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/774).  (\*)
-
-- Move report extension to a task extension to make reports smaller.  (\*)
-
 04:
 
-- Redefine `time_precision` as its own `TimePrecision` type, to maintain
-  compatibility with DAP-16 {{!DAP}}.
+- Move task binding is moved to {{!DAP}}
+  (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/774).  (\*)
+
+- Remove the report extension.  (\*)
 
 03:
 
@@ -146,7 +147,7 @@ this mechanism was used to configure the task.
 
 This document uses the same conventions for error handling as {{!DAP}}. In
 addition, this document extends the core specification by adding the following
-error types:
+DAP error type:
 
 | Type        | Description                                                                               |
 |:------------|:------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap-taskprov.md
```diff
@@ -277,6 +241,23 @@ For example:
     DAP-Taskprov: :AWYAFGh0dHBzOi8vZXhhbXBsZS5jb20vABxodHRwczovL2Fub3RoZXIuZXhhbXBsZS5jb20vAAAAAAAAA+gAAAPoAgAAAAAAAAAAJxAAAAAAAAACXQAAAAIABAAAAP8ACAAAAAQwMTIz:
```
cjpatton, line 236: @tgeoghegan would you mind double checking that this is a proper TaskConfiguration, and if not, re-generating it?

tgeoghegan: Here's a valid encoding of the following task:

```
TaskConfiguration {
    task_info: [ // "test"
        116,
        101,
        115,
        116,
    ],
    leader_aggregator_endpoint: https://leader.example.com/,
    helper_aggregator_endpoint: https://helper.example.com/,
    time_precision: TimePrecision(
        60,
    ),
    min_batch_size: 10,
    batch_config: TimeInterval,
    vdaf_config: Prio3Count,
    extensions: [
        TaskInterval(
            Interval {
                start: Time(
                    60,
                ),
                duration: Duration(
                    100,
                ),
            },
        ),
    ],
}
```

```suggestion
    DAP-Taskprov: :BHRlc3QAG2h0dHBzOi8vbGVhZGVyLmV4YW1wbGUuY29tLwAbaHR0cHM6Ly9oZWxwZXIuZXhhbXBsZS5jb20vAAAAAAAAADwAAAAAAAAACgEAAAAAAAEAAAAUAAEAEAAAAAAAAAA8AAAAAAAAAGQ=:
```

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -277,6 +241,23 @@ For example:
...
 ~~~
 
+## Deriving the DAP Task ID {#dap-task-id}
+
+When the Taskprov mechanism is in use, the task ID  ({{Section 4.2 of !DAP}} )
```
tgeoghegan:
```suggestion
When the Taskprov mechanism is in use, the task ID ({{Section 4.2 of !DAP}} )
```

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -496,45 +465,22 @@ these kinds of attack by:
...
-The task provisioning mechanism in {{taskprov}} is designed so that the
-Aggregators do not need to store individual task configurations long-term.
-Because the task configuration is advertised in each request in the upload,
-aggregation, and collection protocols, the process of opting-in and deriving the
-task ID and VDAF verify key can be re-run on the fly for each request. This is
-useful if a large number of concurrent tasks are expected. Once an Aggregator
-has opted-in to a task, the expectation is that the task is supported until it
-ends. In particular, Aggregators that operate in this manner MUST NOT opt
-out once they have opted in.
+The DAP-Taskprv provisioning mechanism is designed so that the Aggregators do
```
tgeoghegan, line 462:
```suggestion
The DAP-Taskprov provisioning mechanism is designed so that the Aggregators do
```

tgeoghegan (approved):
LGTM, just a couple typos and needs the example fixed.

draft-ietf-ppm-dap-taskprov.md
cjpatton: Thanks!

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -66,11 +68,15 @@ scheme to use for the secure computation (a Verifiable Distributed Aggregation
...
 
-{{taskprov}} specifies one possible mechanism for provisioning DAP tasks that is
-built on top of the task configuration definition in {{Section 4.2 of !DAP}}.
-Its chief design goal is to make task configuration completely in-band, via HTTP
-request headers. It also defines a task extension that signals to parties that
-this mechanism was used to configure the task.
+This document specifies a mechanism for provisioning DAP tasks that is built on
+top of the task configuration definition in {{Section 4.2 of !DAP}}. Its chief
+design goal is to make task configuration completely in-band, via HTTP request
+headers. Long-level cryptographic assets, such as HPKE configurations
```
divergentdave: nit: typo
```suggestion
headers. Long-lived cryptographic assets, such as HPKE configurations
```

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -83,7 +89,7 @@ this mechanism was used to configure the task.
 - Task binding is moved to {{!DAP}}
   (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/774).  (\*)
 
-- Move report extension to a task extension to make reports smaller.  (\*)
+- Remove the report extension.  (\*)
 
 04:
```
divergentdave: This is an outstanding issue dating back to #126: The datatracker has only drafts up to draft-ietf-ppm-dap-taskprov-03, so we should fold these into a single "04 (forthcoming):" header.

draft-ietf-ppm-dap-taskprov.md
```diff
@@ -274,9 +238,26 @@ that are present MUST be ignored.
...
-    DAP-Taskprov: :AWYAFGh0dHBzOi8vZXhhbXBsZS5jb20vABxodHRwczovL2Fub3RoZXIuZXhhbXBsZS5jb20vAAAAAAAAA+gAAAPoAgAAAAAAAAAAJxAAAAAAAAACXQAAAAIABAAAAP8ACAAAAAQwMTIz:
+    DAP-Taskprov: :BHRlc3QAG2h0dHBzOi8vbGVhZGVyLmV4YW1wbGUuY29tLwAbaHR0cHM6Ly9oZWxwZXIuZXhhbXBsZS5jb20vAAAAAAAAADwAAAAAAAAACgEAAAAAAAEAAAAUAAEAEAAAAAAAAAA8AAAAAAAAAGQ=:
 ~~~
 
+## Deriving the DAP Task ID {#dap-task-id}
+
+When the Taskprov mechanism is in use, the task ID ({{Section 4.2 of !DAP}} )
+is set to the hash of the task configuration. This prevents the protocol
+participants from processing reports for the same task using different
+parameters.
```
divergentdave: This sentence isn't correct anymore, since that goal is already accomplished by including those parameters in the `InputShareAad`. Now, this derivation is moreso a convenient way to choose a TaskID in the Taskprov deployment scenario. We could say that this provides a deterministic way for all parties to compute the TaskID from the TaskConfiguration in the headers of the initial request.

divergentdave (approved):

Issue #127, opened 2026-05-29
Specify verification key ID


divergentdave:
In order to work well with the latest drafts of DAP, this document should specify the verification key ID of the verification key derived in section 4.3. The simplest solution would be to just specify that the verification key ID is zero. Alternately, the document could allow for lifecycle management of `verify_key_init` values, and let an identifier of a particular shared secret be the identifier for all verification keys derived from it.

PR #132, opened 2026-07-03
Define the verification key ID

cjpatton:
Closes #127.
Based on #129 (merge that first).

### DIFF
--- a/draft-ietf-ppm-dap-taskprov.md
+++ b/draft-ietf-ppm-dap-taskprov.md
@@ -91,6 +91,8 @@ that satisfies the security requirements for this key ({{Section 9.1 of
 
 - Remove the report extension.  (\*)
 
+- Specify how to pick the VDAF verification key ID as required by DAP.
+
 03:
 
 - Handle repeated extensions in the `TaskprovExtension` field of the
@@ -281,6 +283,9 @@ where `task_id` is as defined in {{dap-task-id}}. Functions HKDF-Extract() and
 HKDF-Expand() are as defined in {{!RFC5869}}. Both functions are instantiated
 with `SHA-256()` as defined in {{SHS}}.
 
+When the VDAF verification key is derived in this way, the verification key ID
+({{Section 4.5.3.1 of !DAP}}) SHALL be `0`.
+
 ## Opting into a Task {#provisioning-a-task}
 
 Prior to participating in a task, each protocol participant must determine if
@@ -362,7 +367,8 @@ request as usual.
 When the Leader opts in to a task, it MUST derive the VDAF verification key
 for that task as described in {{vdaf-verify-key}}. The Leader MUST advertise
 the task to the Helper in every request incident to the task as described in
-{{task-advertisement}}.
+{{task-advertisement}}. The Leader MUST use `0` as the verification key ID when
+initializing an aggregation job ({{Section 4.5.3.1 of !DAP}}).
 
 ### Collection Protocol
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap-taskprov.md
```diff
@@ -400,7 +406,10 @@ Next, the Helper decides whether to opt in to the task as described in
 "invalidTask".
 
 Finally, the Helper completes the request as usual, deriving the VDAF
-verification key for the task as described in {{vdaf-verify-key}}.
+verification key for the task as described in {{vdaf-verify-key}}. The
+verification key ID picked by the Leader for aggregation job initialization
+({{Section 4.5.3.1 of !DAP}}) MUST be `0`; the Helper MUST handle every other
+value as an unrecognized key ID.
```
martinthomson, line 406: Is this last bit necessary?  Is there any need to prohibit this directly.  What if I had a process that added more verification keys to a task?

tgeoghegan: I'm also nervous about forcing the verification key ID to be 0. I guess that some future draft written at PPM that spells out a mechanism for rotating verification keys and using the ID field could update these MUSTs and SHALLs and explain how you use a nonzero verification key ID with taskprov.

tgeoghegan (commented):
This change does what it sets out to do, but let's make sure we agree that taskprov could ever use a nonzero verification key ID before we merge.

draft-ietf-ppm-dap-taskprov.md
cjpatton: Per https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov/pull/130, taskprov is intentionally highly opinionated here: you derive a per-**task** key from a pre-shared secret; there is no such thing as multiple keys per task. This being the case, there's no reason to derive the key ID any other way.

Unless we want to let the key ID to identify the pre-shared secret?

tgeoghegan: Yeah, maybe you could do that. But let's not go design the whole mechanism now. For my part, I'd be satisfied with this change right now if we agree that future drafts can amend the MUST and SHALL stuff we write down today.

tgeoghegan (approved):
Approving, because I'm satisfied that Chris has thought through the implications of forcing `verification_key_id = 0` and did it on purpose.

draft-ietf-ppm-dap-taskprov.md
cjpatton: To be interoperable, I think we would want to add some versioning to the DAP-Taskprov header to allow the client to signal expectations.

divergentdave: I think defining behavior for verification key ID 0 only for now would be fine. However, if extensibility concerns are a barrier for this PR, maybe we should consider bumping up the cardinality of `verification_key_init`. The spec diff may be similar to that we have now for ID 0 only.

My initial thought for how to handle multiple keys would be sharing multiple `verification_key_init` values out of band between aggregators, each with an ID, and then deriving verification keys for tasks from each `verification_key_init` value, assigning the `verification_key_init` value's ID as the verification key ID for corresponding keys derived from it.

cjpatton: Yeah, that's basically what I meant here:

> Unless we want to let the key ID to identify the pre-shared secret?

I'll implement this when I get a chance.

markblunk: Should we specify the error that the Helper responds with if the key is not 0?

PR #135, opened 2026-07-14
MUST opt out if task has unrecognized extension

divergentdave:
DAP says that "Extensions are mandatory to support. Protocol participants MUST NOT participate in tasks containing unrecognized extensions." Thus, I think we should add this to the list of situations where opting out of a task is required.


tgeoghegan (approved):
Wouldn't all the conditions enumerated in draft-ietf-ppm-dap apply implicitly? There's no harm in reminding readers of this requirement, though.

cjpatton (approved):
I think this is already implicit, but it certainly doesn't hurt. We just need to be consistent about reiterating all of the checks and not just some of them.

